PR #42, opened 2017-10-11
*: provide a storage interface

vbatts:
Based on the memstore sample, provide a golang interface for how to
interact with a storage implementation.

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/server-go/storage.go
+++ b/server-go/storage.go
@@ -1,0 +1,60 @@
+package server
+
+import (
+	"github.com/grafeas/grafeas/samples/server/go-server/api"
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/errors"
+)
+
+// Storager is the interface that a Grafeas storage implementation would provide
+type Storager interface {
+	// CreateNote adds the specified note
+	CreateNote(n *swagger.Note) *errors.AppError
+
+	// CreateOccurrence adds the specified occurrence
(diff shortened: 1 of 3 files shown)


server-go/storage.go
```diff
@@ -0,0 +1,61 @@
+package server
+
+import (
+	"github.com/grafeas/grafeas/samples/server/go-server/api"
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/errors"
+)
+
+// Storager implementations are
```
R2wenD2: it looks like this comment didn't get finished?

vbatts:
@R2wenD2 what are these merge commits added to my branch? can i just rebase on master?

server-go/storage.go
vbatts: updated

vbatts:
updated. PTAL

R2wenD2:
Yeah, go ahead and rebase. We renamed a file, I was trying to avoid it impacting you too much :)

vbatts:
I can work on `errors` next. Ideally, a top-level won't import from `./samples` :-)

R2wenD2:
That sounds good. Thanks for the cleanup :0) I'd like to rename the swagger package and hoist that out of samples too.

vbatts:
I was going to look at that too, but need to get familiar with that swagger
codegen process.

PR #46, opened 2017-10-12
*: flatten the samples and generate golang API

vbatts:
this is an invasive patch, but puts a make target on the golang api
generation. With that, the directory structure of the sample can flatten
a bit (and have shorter import paths ;-)

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>


server-go/api/api_response.go
```diff
@@ -0,0 +1,44 @@
...
+ * Generated by: https://github.com/swagger-api/swagger-codegen.git
+ */
+
+package api
+
+import (
+	"net/http"
+)
+
+type APIResponse struct {
```
R2wenD2: We don't need the APIResponse/APIClient/Configuration here, since what we really want is a model for the server.

Can you remove these and add them to the .swagger-codegen-ingore? That should also remove the resty dep which is causing the build to fail.

Makefile
```diff
@@ -18,9 +19,13 @@ fmt:
...
 # http://godoc.org/code.google.com/p/go.tools/cmd/vet
 # go get code.google.com/p/go.tools/cmd/vet
 vet:
 	go vet ./...
 
+$(CODEGEN_JAR):
+	wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O $(CODEGEN_JAR)
+
+server-api: ./v1alpha1/grafeas.json $(CODEGEN_JAR)
+	java -jar $(CODEGEN_JAR) generate -l go -i ./v1alpha1/grafeas.json -o ./server-go/api  -D packageName=api
```
R2wenD2: We probably want to version this - so that we can support new versions of the model in the future.

server-go/api/api_response.go
vbatts: I don't see that it is ignored in master. :-\

Makefile
vbatts: right. I'll make this `./server-go/api/v1alpha1`

server-go/api/api_response.go
vbatts: also, what did you use to do the vendor.json? not glide, godep or dep

R2wenD2: govendor

I _should_ have ignored in master. I just deleted the files initially, I planned to come back to it. You got there first :)

vbatts: oh! that makes sense. The `.swagger-codegen-ignore` seems to ignore patterns of file paths, not of structs to _not_ be rendered. I will try my hand at this. And investigate govendor for future use. Shame it leaks your local paths :-\

vbatts:
ok. rebased on master. versioned the api. and added vendor things, as the manual thing you did seems not straight forward to automate.

Now i'm investigating why the `Package` field is giving issues.

vbatts:
ok. got it.

vbatts:
oh gross. that makes ./vendor huge. :grimacing:

vbatts:
i'm realizing that this overlaps entirely with the client-go repo. That was not intentional, and I'm not sure how you'd like to proceed.

vbatts:
found and opened https://github.com/swagger-api/swagger-codegen/pull/6669 in the course of this

vbatts:
Closing as we're moving away from the swagger/OpenAPI, to protobuf/grpc instead

PR #66, opened 2017-11-13
Fixing output path in Make file and adding a Git ignore.

vtsao:


Makefile
```diff
@@ -30,7 +30,7 @@ model-go/v1alpha1/v1alpha1/proto/grafeas.pb.go: .install.protoc-gen-go v1alpha1/
 		-I. \
 		-I ./vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
 		-I ./vendor/github.com/googleapis/googleapis \
-		--go_out=./v1alpha1/ \
+		--go_out=./ \
```
vbatts: interesting. For me, this puts the output in the wrong directory.

vtsao: Which directory does it put it in for you? What version of protoc are you using? For me it puts it in `v1alpha1/v1alpha1/proto`, which is incorrect.

The [go_out](https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation) flag replaces the proto path ("./") portion of the proto file (v1alpha1/proto/grafeas.proto) with the go_out value. And this would yield: v1alpha1/v1alpha1/proto/grafeas.proto.

vbatts: interesting. It _had_ put is in ./proto/, but not it does not? :grimacing: 

```
$ protoc --version
libprotoc 3.2.0
```
It seems to do the right thing now. Thanks!

PR #80, opened 2017-12-12
KubeCon demo changes to example Grafeas server

lizrice:
Minor changes to the example server that I needed for my demo. I won't be offended if you don't take them as they aren't exactly "complete"! But hopefully a very small step in the right direction. 

- On createNote, the Google container analysis API requires there not to be a name provided in the note, so I made this compatible. 

- Primitive filtering for listOccurrences (only). 

- Removed duplicate settings of statusOK


lizrice:
I think this is no longer relevant after the move to grpc, so let's close this. We'll need to do some updates for grpc on our side, and if it turns out similar changes to these are still needed on the server side we can revisit.

PR #81, opened 2017-12-13
Project crud

### LINKED ISSUE #40: Add CRUD for projectIds to the api

furuholm:
Fixes #40


samples/server/go-server/api/server/storage/memstore_test.go
```diff
@@ -105,17 +146,35 @@ func TestDeleteOccurrence(t *testing.T) {
 	}
 }
 
+func TestDeleteProject(t *testing.T) {
+	s := NewMemStore()
+	pID := "myproject"
+	// Delete before the note exists
```
R2wenD2: s/note/project

samples/server/go-server/api/server/testing/testobjects.go
```diff
@@ -15,16 +15,18 @@
...
 	"github.com/golang/protobuf/ptypes/any"
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/name"
 	pb "github.com/grafeas/grafeas/v1alpha1/proto"
 	opspb "google.golang.org/genproto/googleapis/longrunning"
-	"log"
 )
 
-func Occurrence(noteName string) *pb.Occurrence {
-	return &pb.Occurrence{
+func Occurrence(noteName string) (*pb.Occurrence, string) {
```
R2wenD2: I think it would be better to actually have these take a project as an argument so that future tests can easily create objects in separate projects.

v1alpha1/proto/grafeas.proto
```diff
@@ -1250,6 +1285,29 @@ message ProjectRepoId {
 // there would be one `Note` for the vulnerability and an `Occurrence` for
 // each package with the vulnerability referencing that `Note`.
 service Grafeas {
+  // Creates a new `Project`.
+  rpc CreateProject(CreateProjectRequest) returns (google.protobuf.Empty) {
```
R2wenD2: I'd prefer the Project CRUD in a separate service, as I can imagine people having a preexisting project or account structure in place, which may be outside of this.

samples/server/go-server/api/server/testing/testobjects.go
furuholm: Sounds good. Same for notes and operations then I guess?

v1alpha1/proto/grafeas.proto
furuholm: Ok. And then we make the location of the project CRUD service becomes a configuration parameter? 
Name of service? Projects? GrafeasProjects?

R2wenD2: Yes on project CRUD being a config parameter. 

GrafeasProjects sgtm, it avoids namespace duplication.

samples/server/go-server/api/server/storage/memstore_test.go
furuholm: Fixed in d888a9e

samples/server/go-server/api/server/testing/testobjects.go
furuholm: Proposed fix in 1c1be49

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -31,12 +32,49 @@ type memStore struct {
...
+// DeleteProject deletes the project with the given pID from the mem store
+func (m *memStore) DeleteProject(pID string) error {
+	if _, ok := m.projects[pID]; !ok {
+		return status.Error(codes.InvalidArgument, fmt.Sprintf("Project with name %q does not Exist", pID))
+	}
+	delete(m.projects, pID)
+	return nil
+}
+
+func (m *memStore) GetProject(pID string) (*pb.Project, error) {
```
R2wenD2: Can you add comment for this (its a public method).

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -31,12 +32,49 @@ type memStore struct {
...
+	if _, ok := m.projects[pID]; !ok {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Project with name %q does not Exist", pID))
+	}
+	return &pb.Project{ProjectId: pID}, nil
+}
+
+func (m *memStore) ListProjects(filters string) []string {
```
R2wenD2: Please adda  comment here.

samples/server/go-server/api/server/testing/testobjects.go
```diff
@@ -15,17 +15,20 @@
...
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
 	pb "github.com/grafeas/grafeas/v1alpha1/proto"
 	opspb "google.golang.org/genproto/googleapis/longrunning"
-	"log"
 )
 
-func Occurrence(noteName string) *pb.Occurrence {
+func Occurrence(pID, noteName string) *pb.Occurrence {
+	name := fmt.Sprintf("projects/%s/occurrences/134", pID)
```
R2wenD2: Can you inline this below, since I think it is only used once?

samples/server/go-server/api/server/testing/testobjects.go
```diff
@@ -59,9 +62,10 @@ func Occurrence(noteName string) *pb.Occurrence {
 	}
 }
 
-func Note() *pb.Note {
+func Note(pID string) *pb.Note {
+	name := fmt.Sprintf("projects/%s/notes/CVE-1999-0710", pID)
```
R2wenD2: Same here, I think you can incline this

samples/server/go-server/api/server/testing/testobjects.go
```diff
@@ -165,15 +169,16 @@ func Note() *pb.Note {
...
 
-func Operation() *opspb.Operation {
+func Operation(pID string) *opspb.Operation {
 	md := &pb.OperationMetadata{CreateTime: ptypes.TimestampNow()}
 	bytes, err := proto.Marshal(md)
 	if err != nil {
 		log.Printf("Error parsing bytes: %v", err)
 		return nil
 	}
+	name := fmt.Sprintf("projects/%s/operations/foo", pID)
```
R2wenD2: Inline please

samples/server/go-server/api/server/v1alpha1/impl.go
```diff
@@ -313,3 +335,35 @@ func (g *Grafeas) ListNoteOccurrences(ctx context.Context, req *pb.ListNoteOccur
 func (g *Grafeas) CancelOperation(context.Context, *opspb.CancelOperationRequest) (*empty.Empty, error) {
 	return nil, status.Error(codes.Unimplemented, "Currently Unimplemented")
 }
+
+// GrafeasProjects is an implementation of the parts of Grafeas API that manages projects.
+type GrafeasProjects struct {
+	S server.Storager
+}
+
+// CreateNote validates that a note is valid and then creates a note in the backing datastore.
```
R2wenD2: s/CreateNote/CreateProject
s/note/project

samples/server/go-server/api/server/v1alpha1/impl.go
```diff
@@ -313,3 +335,35 @@ func (g *Grafeas) ListNoteOccurrences(ctx context.Context, req *pb.ListNoteOccur
...
+func (g *GrafeasProjects) CreateProject(ctx context.Context, req *pb.CreateProjectRequest) (*empty.Empty, error) {
+	projectId := req.ProjectId
+	if projectId == "" {
+		log.Printf("Invalid project id: %v", projectId)
+		return &empty.Empty{}, status.Error(codes.InvalidArgument, "Invalid project id")
+	}
+	return &empty.Empty{}, g.S.CreateProject(projectId)
+}
+
+func (g *GrafeasProjects) ListProjects(ctx context.Context, req *pb.ListProjectsRequest) (*pb.ListProjectsResponse, error) {
```
R2wenD2: Please add a function comment here.

server-go/storage.go
```diff
@@ -21,6 +21,9 @@ import (
 
 // Storager is the interface that a Grafeas storage implementation would provide
 type Storager interface {
+	// CreateProejct adds the specified project
```
R2wenD2: CreateProject

server-go/storage.go
```diff
@@ -30,6 +33,9 @@ type Storager interface {
 	// CreateOperation adds the specified operation
 	CreateOperation(o *opspb.Operation) error
 
+	// DeleteNote deletes the project with the given pID
```
R2wenD2: DeleteProject

v1alpha1/proto/grafeas.proto
```diff
@@ -23,7 +23,36 @@ import "google/longrunning/operations.proto";
...
+
+// Request to get a Project.
+message GetProjectRequest {
+  string project_id = 1;
+}
+
+// Request to list projects.
+message ListProjectsRequest {
+  // The filter expression.
+  string filter = 2;
```
R2wenD2: No need to start with 2 here - we can start with 1.

v1alpha1/proto/grafeas.proto
```diff
@@ -1365,3 +1410,35 @@ service Grafeas {
     };
   }
 }
+
+service GrafeasProjects {
```
R2wenD2: Can you add a top level service comment?

v1alpha1/proto/grafeas.proto
```diff
@@ -1365,3 +1410,35 @@ service Grafeas {
...
+  // Creates a new `Project`.
+  rpc CreateProject(CreateProjectRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      post:  "/v1alpha1/projects/{project_id=*}"
```
R2wenD2: Can you change this to be /v1alpha1/{name=projects/*} to be consistent with the other api definitions? here and below.

v1alpha1/proto/grafeas.proto
furuholm: I've put GrafeasProjects in its own service which I inject into the Grafeas service. Was this about what you had in mind?
It's not finished though.
1. For now I'm using the same storage object for Grafeas and GrafeasProjects. If they should be separated I think the Storager interface should be split as well.
2. The Grafeas service still invokes the GrafeasProject methods using a local object rather than a gRPC client. Not sure gRPC always should be used or if we only call out when projects are located somewhere else. Thoughts?
3. Configuration of GrafeasProjects service location still remains (as a consequence of 2)
4. I realized that I hade put the "make sure project exist" check in the wrong layer before. I had put it in the storage layer. Now it is in the application layer instead (impl.go).

R2wenD2: GrafeasProjects looks good. 

1. I pondered this when reviewing, I think it makes sense to split Storager, its easy to implement both interfaces in the same package if one wants to - however I can also imagine calling out to a different backend for project info.
2. I think for now we can keep it simple, and use the local object. If we need to split it up later we can. We're running both in the same service now so that's fine.
3. GrafeasProjects can just be added to the existing server. If we want to break it up later we can, but no need to over engineer that now :)
4. Yeah - that looks good.

furuholm: Ok. I'll split the interface. Makes sense to split projects from the rest all the way down.

samples/server/go-server/api/server/testing/testobjects.go
furuholm: Fixed in 11c4c01

samples/server/go-server/api/server/testing/testobjects.go
furuholm: Fixed in 901f777

v1alpha1/proto/grafeas.proto
furuholm: Yeah. Was going back and forth on this. You are probably right though. Will fix this tomorrow.

furuholm:
Comments should be fixed now as well. 

I will fix the project routes tomorrow and also switch to return Project:s instead of strings from ListProjects to be consistent with GetProject.

v1alpha1/proto/grafeas.proto
furuholm: Please have an extra look at this one. I think there is room for improvement.

v1alpha1/proto/grafeas.proto
furuholm: Fixed in 715d9e0

furuholm:
I think I've fixed everything you found so far @R2wenD2. I've also changed from strings to Projects in ListProjects and split Storager. 
(BTW: why is it called Storager and not just Storage?)

R2wenD2:
@furuholm I didn't get a chance to review this today, and I'm going to be away until 1/3. @skelterjohn can help with review while I'm away.

samples/server/go-server/api/server/storage/memstore_test.go
```diff
@@ -29,7 +30,10 @@ import (
 
 func TestCreateNote(t *testing.T) {
 	s := NewMemStore()
-	n := testutil.Note()
+	ps := NewProjectMemStore()
```
skelterjohn: What is the purpose of creating ps and calling ps.CreateProject()? It doesn't appear that the function affects any global state that influences the test...

samples/server/go-server/api/server/storage/project_memstore.go
```diff
@@ -0,0 +1,72 @@
...
+
+// NewMemStore creates a memStore with all maps initialized.
+func NewProjectMemStore() server.ProjectStorager {
+	return &projectMemStore{make(map[string]bool)}
+}
+
+// CreateProject adds the specified project to the mem store
+func (m *projectMemStore) CreateProject(pID string) error {
+	if _, ok := m.projects[pID]; ok {
+		return status.Error(codes.InvalidArgument, fmt.Sprintf("Project with name %q already exists", pID))
```
skelterjohn: `codes.AlreadyExists`. Use `codes.InvalidArgument` for data that is improperly formatted or makes no sense.

samples/server/go-server/api/server/storage/project_memstore.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/name"
+	"github.com/grafeas/grafeas/server-go"
+	pb "github.com/grafeas/grafeas/v1alpha1/proto"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// projectMemStore is an in-memory storage solution for Grafeas
+type projectMemStore struct {
```
skelterjohn: Protect the `.projects` field with a mutex. Otherwise you have to protect it from the caller's side which is much harder to manage (esp between multiple callers).

samples/server/go-server/api/server/storage/project_memstore.go
```diff
@@ -0,0 +1,72 @@
...
+		return status.Error(codes.InvalidArgument, fmt.Sprintf("Project with name %q does not Exist", pID))
+	}
+	delete(m.projects, pID)
+	return nil
+}
+
+// GetProject returns the project with the given pID from the mem store
+func (m *projectMemStore) GetProject(pID string) (*pb.Project, error) {
+	if _, ok := m.projects[pID]; !ok {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Project with name %q does not Exist", pID))
```
skelterjohn: `codes.NotFound`

samples/server/go-server/api/server/v1alpha1/impl.go
```diff
@@ -43,9 +44,19 @@ func (g *Grafeas) CreateNote(ctx context.Context, req *pb.CreateNoteRequest) (*p
...
+		return nil, status.Error(codes.InvalidArgument, "Note name must not be empty")
+	}
+	pID, _, err := name.ParseNote(n.Name)
+	if err != nil {
 		log.Printf("Invalid note name: %v", n.Name)
 		return nil, status.Error(codes.InvalidArgument, "Invalid note name")
 	}
+	if _, err = g.Projects.GetProject(ctx, &pb.GetProjectRequest{Name: name.FormatProject(pID)}); err != nil {
+		log.Printf("Unable to get project %v, err: %v", pID, err)
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Project %v not found", pID))
```
skelterjohn: `codes.NotFound`

samples/server/go-server/api/server/v1alpha1/impl.go
```diff
@@ -84,6 +100,11 @@ func (g *Grafeas) CreateOperation(ctx context.Context, req *pb.CreateOperationRe
 		log.Printf("Invalid operation name: %v", o.Name)
 		return nil, status.Error(codes.InvalidArgument, "Invalid operation name")
 	}
+	pID, _, err := name.ParseOperation(o.Name)
+	if _, err = g.Projects.GetProject(ctx, &pb.GetProjectRequest{Name: name.FormatProject(pID)}); err != nil {
+		log.Printf("Unable to get project %v, err: %v", pID, err)
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Project %v not found", pID))
```
skelterjohn: `codes.NotFound`

samples/server/go-server/api/server/storage/memstore_test.go
furuholm: Ahhh...good catch! Previous versions required there to be a project to be able to create notes/occurrences/operations. This is now enforced in the application layer instead.

Fixed in 7e350b5

samples/server/go-server/api/server/storage/project_memstore.go
furuholm: Agree! Fixed in 9f66111. Not sure my solution is best practice so please take an extra look at this.

samples/server/go-server/api/server/storage/project_memstore.go
furuholm: Fixed in 55f8d67

samples/server/go-server/api/server/v1alpha1/impl.go
furuholm: Fixed in 55f8d67

furuholm: Should NotFound be returned even if the Project is missing? One could argue that NotFound should refer to the resource that was provided by the user (in this case the note). I think returning NotFound would make sense as well. Just want to double check.

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -162,6 +186,8 @@ func (m *memStore) ListNoteOccurrences(pID, nID, filters string) ([]*pb.Occurren
 	}
 	nName := name.FormatNote(pID, nID)
 	os := []*pb.Occurrence{}
+	m.RLock()
```
furuholm: There is a potential race here if another user deletes note between GetNote and m.RLock(). I wasn't sure how to properly do this without risking deadlocking. After reading up on RLock it seems to me that it should be ok to RLock before GetNote, whereas Lock would cause a deadlock.

Does RLocking before GetNote look ok to you @skelterjohn?

furuholm: Made the change in 4627a80. Let me know if you disagree.

samples/server/go-server/api/server/storage/project_memstore.go
```diff
@@ -0,0 +1,82 @@
...
+
+// projectMemStore is an in-memory storage solution for Grafeas
+type projectMemStore struct {
+	sync.RWMutex
+	projects map[string]bool
+}
+
+// NewMemStore creates a memStore with all maps initialized.
+func NewProjectMemStore() server.ProjectStorager {
+	return &projectMemStore{sync.RWMutex{}, make(map[string]bool)}
```
skelterjohn: Same comment w/ not needing to initialize the mutex and named fields for the struct literal.

samples/server/go-server/api/server/v1alpha1/impl.go
skelterjohn: NotFound is what makes most sense to me. InvalidArgument should be reserved for things that never make sense - not for things that don't work with the current state of the world (eg that project doesn't exist, or the user doesn't have access).

server-go/project_storage.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	pb "github.com/grafeas/grafeas/v1alpha1/proto"
+)
+
+// ProjectStorager is the interface that a Grafeas storage implementation would provide
+// for Project enteties
+type ProjectStorager interface {
+	// CreateProject adds the specified project
+	CreateProject(pID string) error
```
skelterjohn: these interface methods should have a context as the first parameter. With context, you go all in. Annoying to plumb through, but the idea is that perhaps the underlying .CreateProject() call also makes an RPC to write the data - it would want to time out on the parent context as well.

samples/server/go-server/api/server/v1alpha1/impl.go
furuholm: Agree. Fixed in 153e507.

samples/server/go-server/api/server/v1alpha1/impl.go
furuholm: Fixed in 153e507.

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -28,21 +29,24 @@ import (
...
+	sync.RWMutex
 	occurrencesByID map[string]*pb.Occurrence
 	notesByID       map[string]*pb.Note
 	opsByID         map[string]*opspb.Operation
 }
 
 // NewMemStore creates a memStore with all maps initialized.
 func NewMemStore() server.Storager {
-	return &memStore{make(map[string]*pb.Occurrence), make(map[string]*pb.Note),
+	return &memStore{sync.RWMutex{}, make(map[string]*pb.Occurrence), make(map[string]*pb.Note),
```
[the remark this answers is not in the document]
furuholm: Fixed in e4f424a.

samples/server/go-server/api/server/storage/project_memstore.go
furuholm: Fixed in e4f424a.

server-go/project_storage.go
furuholm: Sounds like a good idea but I don't think it is within scope of this PR as this would require corresponding changes to `Storager` as well. Would also like @R2wenD2  to approve of this before changing the interface. Will open a new issue for this instead.

v1alpha1/proto/grafeas.proto
```diff
@@ -1365,3 +1416,39 @@ service Grafeas {
     };
   }
 }
+
+// [GrafeasProjects](grafeas.io) API.
+//
+// Manages Grafeas `Projects`. `Projects` contain sets of other Grafeas
```
vbatts: this line has trailing whitespace

vbatts:
so many commits! are they all needed or might could be squashed a bit?

furuholm:
@vbatts Sure I can do that. How do I do it best without messing up the code review? Just squash to a single commit? Create new PR?

furuholm:
@vbatts created a new branch with everything in a single commit. Hope that helps. https://github.com/combient/Grafeas/tree/project-crud-squashed

furuholm:
Agree with @vbatts that this PR is getting out of control. Created PR #87 with squashed history to make things easier.

PR #87, opened 2018-01-03
Add project CRUD to gRPC API

furuholm:
Replaces PR #81. Squashed history to make it easier to review.


samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -28,21 +29,74 @@ import (
...
+}
+
+// ListProjects returns the project id for all projects from the mem store
+func (m *memStore) ListProjects(filters string) []*pb.Project {
+	m.RLock()
+	defer m.RUnlock()
+	projects := make([]*pb.Project, len(m.projects))
+	i := 0
+	for k := range m.projects {
+		projects[i] = &pb.Project{Name: name.FormatProject(k)}
```
skelterjohn: `i` and `k` always have the same values here.

furuholm: I think that would be the case if m.Projects was an array, but m.Projects is a `map[string]bool` (to emulate a set).

skelterjohn: whoops, sorry. saw `projects :=` right above and I lost track of which was which.

R2wenD2:
@furuholm I'll merge this tomorrow morning if you update/merge conflicts.

furuholm:
@R2wenD2 Thanks! I have rebased on top of master.

Issue #84, opened 2017-12-19
Add database storage backend


furuholm:
A database will be required to run Grafeas in anything but development mode.

R2wenD2:
We would get a lot of flexibility for querying various metadata by using Elastic (https://github.com/elastic/elasticsearch) in development go client: https://github.com/elastic/go-elasticsearch

furuholm:
Sounds like a good idea.

Any ideas on type of database? Normalizating all those nested fields will be messy and require lots of join ops during querying so if we go the SQL route (which seems reasonable to me) it’s probably a good idea to store some of the data in serialized form.

R2wenD2:
PostgreSQL would be a good option, as we could use the built in [text search] (https://www.postgresql.org/docs/current/static/textsearch-features.html) if we parse incoming protos into documents. 

We could then store fields required for get/indexing as well as the bytes of the proto, while indexing the proto into a ts_vector in our table.

furuholm:
Maybe the [JSON support in PostgreSQL](https://www.postgresql.org/docs/9.6/static/datatype-json.html) could be of interest as well? I have not used it myself (and I'm not sure how it interacts with the text search features) so it would be interesting to hear what others have to say about it.

furuholm:
Is the strategy to build a CEL compiler that builds postgresql text search queries? It seems to me that a CEL compiler for JSONB queries should be easier to write and get right than one for text strings as there is more structure in the data to leverage.

furuholm:
Until the details have been settled, would a PR with a basic PostgreSQL implementation of the Storager interface be of interest or are there other plans for making this happen? If there is interest I could take this on.

I suggest we start simple with one table each for projects, notes, occurrences and operations. The tables only have to contain name and marshaled data to start with. Getting this and the infrastructure in place would set us up for continued work.

A couple of notes:
- Since there will be more than one type of storage there needs to be a way to select which one to use. I personally like how [Clair](https://github.com/coreos/clair) has strucutured its storage implementation by letting each storage implementation register itself using an init function (see [pgsql.go](https://github.com/coreos/clair/blob/master/database/pgsql/pgsql.go) for an example). I suggest we do something similar.
- There needs to be a config file to configure the database (and other things). I suggest a YAML file.
- I think we could postpone support for database migrations for now.

R2wenD2:
[...]

PR #88, opened 2018-01-05
Update sample server to accept json requests as well

R2wenD2:
This completes server side grpc work. The various clients will need updates to actually work properly with the new server. 

I've also added insecure certs to communicate between the servers. Those should be updated to not be hard coded.


furuholm:
Tried `curl 'http://localhost:10000/v1alpha1/notes'` but got `http: TLS handshake error from [::1]:51176: tls: first record does not look like a TLS handshake` in the server console. 

`curl 'https://localhost:10000/v1alpha1/notes'` causes `http: TLS handshake error from [::1]:51198: remote error: tls: unknown certificate authority`.

Are the insecure certs just a placeholder or do they serve some other purpose? Maybe this should be solved once we have a working config system to allow users to provide their own PKI configuration? (I have added configurability as part of the [branch](https://github.com/combient/Grafeas/tree/postgres-storage) that addresses #84. This could be backported if you don't want to wait for this PR to be merged.)

samples/server/go-server/api/server/cert/certs.go
```diff
@@ -0,0 +1,96 @@
+// cert is an provides insecure keys.
```
furuholm: cert provides insecure keys?

R2wenD2:
@furuholm - you need to pass certs with curl or run curl -k in order for this to work. 

The certs are just a placeholder now. As a follow on to this pr, I'd prefer that we have flags to point to the cert files (which can then be part of our config).

samples/server/go-server/api/server/cert/certs.go
R2wenD2: Fixed.

samples/server/go-server/api/server/main/main.go
```diff
@@ -15,34 +15,87 @@
...
 	flag.Parse()
+	grafeasEndpoint := fmt.Sprintf("localhost:%d", *port)
+
+	opts := []grpc.ServerOption{
+		grpc.Creds(credentials.NewClientTLSFromCert(cert.Pool(), grafeasEndpoint))}
+	grpcServer := grpc.NewServer(opts...)
+	g := v1alpha1.Grafeas{S: storage.NewMemStore()}
+	pb.RegisterGrafeasServer(grpcServer, &g)
+	opspb.RegisterOperationsServer(grpcServer, &g)
 	log.Printf("Server started on port %v", *port)
```
skelterjohn: this log message will only be true after line 78

PR #89, opened 2018-01-09
In toto POC

R2wenD2:
This is a POC for integrating in-toto with Grafeas. Comments welcome, just know that  I don't plan to merge this without an actual design doc and plans for usage.


lukpueh:
This is great and thanks for the initiative! I saw that there are some deviations from the in-toto metadata specification in naming and structure, but I'm sure we can fix these easily once we have a better understanding of how grafeas and in-toto should work together.

We also need to discuss the mapping of links/occurrences and the corresponding layouts/notes. I saw that you added a `layout_note` identifier to the occurrence representation of in-toto links.
In our current design we don't associate links with layouts directly, but instead with the individual steps listed in a layout.

Another important thing that we need to keep an eye on are signatures on the link and layout metadata.

In any case, I'm excited about seeing things move forward. :)

R2wenD2:
Thanks for taking a look @lukpueh its probably easier to chat in person about what we can do to make this better :)

v1alpha1/proto/grafeas.proto
```diff
@@ -601,6 +601,53 @@ message Discovery {
...
+
+message Inspect {
+  string name = 1;
+  repeated InputOutput expected_input = 2;
+  repeated InputOutput expected_output = 3;
+  string run = 4;
+}
+
+// format for materials and products
+message InputOutput {
```
lukpueh: I see you are using `InputOutput` for `expected_materials` and `expected_products` in  `Step` and `Inspect` as well as for `materials` and `products` in `IntotoVerification` (i.e. in-toto links). However, these are different things. 

The fields on Steps and Inspections should be lists of artifact rules, parsable by [`artifact_rules.unpack_rule`](https://github.com/in-toto/in-toto/blob/0.2.0/in_toto/artifact_rules.py#L24-L38). These rules are used to specify which artifacts (materials or products) a step should work with, e.g.:
-  *step B's material foo must match (by hash) step A's product foo*, or 
- *step C creates artifact bar*

[An artifact rule example can be found in the demo layout](https://github.com/in-toto/in-toto/blob/0.2.0/tests/demo_files/demo.layout.template#L88-L107).

`materials` and `products` in in-toto links on the other hand, provide the signed evidence of the artifacts a step worked with, used when applying above artifact rules.
[Material and product examples can be found in demo link files](https://github.com/in-toto/in-toto/blob/0.2.0/tests/demo_files/package.2f89b927.link#L22-L31).

v1alpha1/proto/grafeas.proto
```diff
@@ -601,6 +601,53 @@ message Discovery {
...
+  string file_path = 1;
+  // TODO: should this be an enum?
+  string hash_algorithm = 2;
+  string file_hash = 3;
+}
+
+message KeyDetail {
```
lukpueh: The format of the layout's `keys` field is defined in [`in_toto.formats. ANY_PUBKEY_DICT_SCHEMA`](https://github.com/in-toto/in-toto/blob/0.2.0/in_toto/formats.py#L45-L47), i.e.:
a dictionary with `keyid`s as dictionary keys and public keys that match either [`securesystemslib.formats.PUBLIC_KEY_SCHEMA`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.10.9/securesystemslib/formats.py#L241-L246) or [`in_toto.gpg.formats.PUBKEY_SCHEMA`](https://github.com/in-toto/in-toto/blob/0.2.0/in_toto/gpg/formats.py#L103-L104) as values.

v1alpha1/proto/grafeas.proto
```diff
@@ -762,6 +817,35 @@ message Occurrence {
...
+  // Describes how this occurrence can be verified
+  oneof verification {
+    IntotoVerification intoto = 16;
+  }
+}
+
+// Includes information about how to verify an occurrence using in-toto
+// TODO(r2wend2): This copies some information from the occurrence itself need to reconcile before merging.
+message IntotoVerification {
+  string layout_note = 1;
```
lukpueh: As mentioned in my original [review comment]( https://github.com/grafeas/grafeas/pull/89#issuecomment-356754636), we currently associate link metadata and layout metadata indirectly by step names, i.e. [upon final product verification](https://github.com/in-toto/in-toto/blob/0.2.0/in_toto/verifylib.py#L1289-L1293),  for each step in a the layout to be verified, we try to load `<step.threshold>` links identified by filenames `<step.name>.<authorized functionary keyid prefix>.link`.

We don't want to associate links directly with layouts, because a link does not need to know the layout it belongs to at creation time. Also links should be available for reuse with different layouts.

v1alpha1/proto/grafeas.proto
```diff
@@ -762,6 +817,35 @@ message Occurrence {
...
+  repeated InputOutput materials = 2;
+  repeated InputOutput products = 3;
+  repeated ByProduct byproducts = 4;
+  string command = 5;
```
lukpueh: See comment about `expected_command` datatype above.

v1alpha1/proto/grafeas.proto
```diff
@@ -762,6 +817,35 @@ message Occurrence {
...
   google.protobuf.Timestamp update_time = 10;
+
+  // Describes how this occurrence can be verified
+  oneof verification {
+    IntotoVerification intoto = 16;
+  }
+}
+
+// Includes information about how to verify an occurrence using in-toto
+// TODO(r2wend2): This copies some information from the occurrence itself need to reconcile before merging.
```
lukpueh: Maybe I'm reading this comment wrong, but I wouldn't say that in-toto links (AFAICS described by this message) include information about how to verify in-toto occurrences (layouts). They actually provide the required signed evidence (which functionary, used command, materials, products, environment, ...) in order to verify that the supply chain was carried as specified in the corresponding in-toto layout.

R2wenD2:
after discussion with the in-toto team, this prototype isn't quite right. Further prototyping will be done on the in-toto branch.

PR #91, opened 2018-01-11
Configurable TLS

furuholm:
In order for us to continue to use Grafeas we need support for insecure connections now and tls later. I therefore added support for both through a config file. 

NOTE: The code is in large parts taken from Clair as they had refined the concept from the examples linked from the gRPC gateway project quite a lot. If you don't think that this is ok we'll have to start over with this.

I have verified that the TLS configuration works with curl as well as a go client. Curl needs a combined PEM file (see [gist](https://gist.github.com/mtigas/952344)) and `-k`  to work with self signed certificates.


furuholm:
**Note to self and others about generating certificates and accessing the service**

To generate ca, keys and certs I basically followed https://gist.github.com/mtigas/952344 with some tweaks

```
# Create CA
openssl genrsa -out ca.key 2048
# make sure to set Common Name to your domain, e.g. localhost (without port)
openssl req -new -x509 -days 365 -key ca.key -out ca.crt

# Create the Client Key and CSR
openssl genrsa -out client.key 2048
openssl req -new -key client.key -out client.csr

# Create self-signed client cert
openssl x509 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out client.crt

# Convert Client Key to PKCS
openssl pkcs12 -export -clcerts -in client.crt -inkey client.key -out client.p12

# Convert Client Key to (combined) PEM
openssl pkcs12 -in client.p12 -out client.pem -clcerts
```

config.yaml
```
    cafile: ca.crt
    keyfile: ca.key
    certfile: ca.crt
```

curl
```
curl -k --cert path/to/client.pem 'https://localhost:10000/v1alpha1/projects'
```

go-client (adapted from https://gist.github.com/michaljemala/d6f4e01c4834bf47a9c4)
```
package main

import (
	"context"
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"io/ioutil"
	"log"

	pb "github.com/grafeas/grafeas/v1alpha1/proto"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials"
)

var (
	certFile = "/path/to/client.crt"
	keyFile  = "/path/to/client.key"
	caFile   = "/path/to/ca.crt"
)

func main() {
	// Load client cert
	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
	if err != nil {
		log.Fatal(err)
	}

	// Load CA cert
	caCert, err := ioutil.ReadFile(caFile)
	if err != nil {
		log.Fatal(err)
	}
	caCertPool := x509.NewCertPool()
	caCertPool.AppendCertsFromPEM(caCert)

	// Setup HTTPS client
	tlsConfig := &tls.Config{
		Certificates: []tls.Certificate{cert},
		RootCAs:      caCertPool,
	}
	tlsConfig.BuildNameToCertificate()
	creds := credentials.NewTLS(tlsConfig)
	conn, err := grpc.Dial("localhost:10000", grpc.WithTransportCredentials(creds))

	pClient := pb.NewGrafeasClient(conn)

	...
}
```

samples/server/go-server/api/server/config/config.go
```diff
@@ -0,0 +1,59 @@
...
+	Grafeas *Config `yaml:"grafeas"`
+}
+
+// Config is the global configuration for an instance of Grafeas.
+type Config struct {
+	Server *server.Config `yaml:"server"`
+}
+
+// DefaultConfig is a configuration that can be used as a fallback value.
+func DefaultConfig() *Config {
```
R2wenD2: Do we need DefaultConfig to be public?

samples/server/go-server/api/server/config/config.go
```diff
@@ -0,0 +1,59 @@
...
+package config
+
+import (
+	"io/ioutil"
+
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/server"
+	"gopkg.in/yaml.v2"
+)
+
+type File struct {
```
R2wenD2: Please add a comment describing File

samples/server/go-server/api/server/config/config.go
```diff
@@ -0,0 +1,59 @@
...
+	Grafeas *Config `yaml:"grafeas"`
+}
+
+// Config is the global configuration for an instance of Grafeas.
+type Config struct {
```
R2wenD2: I think this also can be private. Its a bit confusing to have a server.Config and a config.Config both public.

samples/server/go-server/api/server/config/config.go
```diff
@@ -0,0 +1,59 @@
...
+		&server.Config{
+			Address:  "localhost:10000",
+			CertFile: "",
+			KeyFile:  "",
+			CAFile:   "",
+		},
+	}
+}
+
+func LoadConfig(fileName string) (*Config, error) {
```
R2wenD2: Please add a comment describing LoadConfig

furuholm:
Great! Will fix tomorrow...

R2wenD2:
Can you also add the great cert generation description to a readme, and link the main readme to it where we reference running the server?

furuholm:
Will do...

furuholm:
I've fixed your comments and added instructions. Feel free to edit them however you like!

It felt weird to add docs on how to create a client using self signed cert without having docs for the insecure case so I added that as well.

It might be wise to make the code in the docs to actual examples to keep the accurate down the line.

furuholm:
I realised once I rebased the postgres branch on top of this that there now are two server-packages (the one I added here and the one in the go-server directory). I have renamed "my" server package api. If it's ok this change will make it in as part of the postgres-PR instead (fixed it during the rebase, rather not backport unless you deem it necessary).

PR #92, opened 2018-01-15
Postgres storage

furuholm:
This PR targets #84. See discussion in #84 for more details.

A couple of notes:
- I have gathered all test cases in a [new file](https://github.com/combient/Grafeas/blob/postgres-storage/samples/server/go-server/api/server/storage/storager_test.go) to avoid test duplication. The test cases have been moved to subtests. [memstore_test.go](https://github.com/combient/Grafeas/blob/postgres-storage/samples/server/go-server/api/server/storage/memstore_test.go) and [pgsqlstore_test.go](https://github.com/combient/Grafeas/blob/postgres-storage/samples/server/go-server/api/server/storage/pgsqlstore_test.go) calls a single function to run all tests. Is there a better way to do this?
- #84 contains an ongoing discussion about the schema that has to be resolved.


furuholm:
Also, this PR does not provide any framework for db migrations. This needs to be added down the line (if not as part of this PR) together with some kind of migration policy that states what kind of data integrity guarantee users should expect when upgrading Grafeas.

samples/server/go-server/api/server/storage/queries.go
```diff
@@ -16,48 +16,48 @@ package storage
 
 const (
 	createTables = `
-		CREATE TABLE IF NOT EXISTS project (
+		CREATE TABLE IF NOT EXISTS projects (
 			id SERIAL PRIMARY KEY,
 			name TEXT NOT NULL UNIQUE
 		);
-		CREATE TABLE IF NOT EXISTS note (
```
R2wenD2: sorry I didn't get back to you sooner on this.

My suggestion is for each of these to separate the project and resource id so we'd have
                        id SERIAL PRIMARY KEY,
                        project_name TEXT NOT NULL UNIQUE
  			note_name TEXT NOT NULL UNIQUE,
  			data TEXT

Where we would interpret project name as only the project_name part of projects/**project_name**
This would allow us to change paths, if we want/need to in the future without storage issues.  There's been some conversation about whether notes should be under a provider path or remain under project path. 

Thoughts?

samples/server/go-server/api/server/storage/queries.go
```diff
@@ -16,48 +16,48 @@ package storage
...
+		CREATE TABLE IF NOT EXISTS notes (
 			id SERIAL PRIMARY KEY,
 			name TEXT NOT NULL UNIQUE,
 			data TEXT
 		);
-		CREATE TABLE IF NOT EXISTS occurrence (
+		CREATE TABLE IF NOT EXISTS occurrences (
```
R2wenD2: It might be easier here for easier querying to pull 

note_project
note_name 
 
into separate columns. This allows things like listNoteOccurrences to be a bit easier to query.

samples/server/go-server/api/server/storage/queries.go
furuholm: Agree. This will be much better.

samples/server/go-server/api/server/storage/queries.go
furuholm: This would change the behaviour. The question is if we want an occurrence to change when its note is changed? 

At this point we serialise the complete occurrence, so unless we change this as well we probably don't want the occurrence to be updated when the note is updated. OTOH, not doing this change could lead to potential data integrity issues.

Sidenote: I personally think we can live with a query that is a bit more complex if it does what we want it to do.

samples/server/go-server/api/server/storage/queries.go
furuholm: Fixed in 90dbb92

samples/server/go-server/api/server/storage/queries.go
furuholm: @R2wenD2 thoughts?

R2wenD2: We want to join against the notes table for the fields that are shared between the note and occurrence, so that we don't need to update *all* occurrences every time a note changes. In the long run, the join should be cheaper than the update.
wdyt?

(also, apologies for the delayed responses).

furuholm: Ok. So then we don't pull `note_project` and `note_name` into separate columns (per initial comment) and instead keep `note_id int REFERENCES notes` and join with notes table to get note_project and note_name. Right? SGTM but goes agains the initial comment.

R2wenD2: I have a mental model where note is keyed on (projectId, note name) and not note_id field. That is definitely not what's in the schema though :) 

Yeah, what you're suggesting works, lets do that.

furuholm: Have a feeling that a higher bandwidth communication channel than a PR review would make it easier to come to a shared understanding on this 😄

Anyway. Since it sounds like the solutions are (more or less) equivalent I'll go ahead with my suggestion. We can always revisit...

furuholm:
No worries :)

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -0,0 +1,492 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/name"
+	pb "github.com/grafeas/grafeas/v1alpha1/proto"
+	"github.com/lib/pq"
+	opspb "google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type PgSQLConfig struct {
```
R2wenD2: What are your thoughts on putting the config and creation into a separate file? 

These are things that aren't going to be modified often in the normal course of development, however the create, insert, delete, list methods will be.

furuholm: Fixed in c551fed

furuholm:
I've split out the config file and rebased the branch on master so if it LGTY we should be good to go...

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -0,0 +1,482 @@
...
+}
+
+func createSourceString(config *PgSQLConfig) string {
+	return fmt.Sprintf("postgres://%s:%s@%s/?sslmode=%s", config.User, config.Password, config.Host, config.SSLMode)
+}
+
+func createSourceStringWithDbName(config *PgSQLConfig) string {
+	return fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=%s", config.User, config.Password, config.Host, config.DbName, config.SSLMode)
+}
+
```
R2wenD2: Can we also put the methods above here into the config file?

furuholm: Done

furuholm:
Thanks!

PR #101, opened 2018-01-26
Use Project instead of string in create project API

furuholm:


furuholm:
@R2wenD2 Could you trigger another build? (I don't have rights to do so). Looks like the db shut down in the middle of the test. Would like to see if it always happens.

PR #102, opened 2018-01-28
Add CORS support to sample app.

ChristopherMacGown:
This adds support for Cross-Origin Resource Sharing to the sample service to make it easier to develop browser-based applications agains the grafeas API.

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -21,6 +21,9 @@ grafeas:
     cafile:
     keyfile:
     certfile:
+    # CORS configuration (optional)
+    cors_allowed_origins:
+    	# - "http://example.net"
   # Supported storage types are "memstore" and "postgres"
   storage_type: "memstore"
   # Postgres options
--- a/docs/running_grafeas.md
+++ b/docs/running_grafeas.md
@@ -157,3 +157,13 @@ func main() {
 	}
 }
 ```
+
+## Enable CORS on the sample server.
+
+### Update config
+
+Add the following to your config file below the `api` key.
+
+    cors_allowed_origins:
+       - "https://some.example.tld"
+       - "https://*.example.net"
--- a/samples/server/go-server/api/server/api/api.go
+++ b/samples/server/go-server/api/server/api/api.go
@@ -29,16 +29,18 @@ import (
 	server "github.com/grafeas/grafeas/server-go"
 	pb "github.com/grafeas/grafeas/v1alpha1/proto"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/rs/cors"
 	opspb "google.golang.org/genproto/googleapis/longrunning"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 )
 
 type Config struct {
-	Address  string `yaml:"address"`  // Endpoint address, e.g. localhost:8080
-	CertFile string `yaml:"certfile"` // A PEM eoncoded certificate file
-	KeyFile  string `yaml:"keyfile"`  // A PEM encoded private key file
-	CAFile   string `yaml:"cafile"`   // A PEM eoncoded CA's certificate file
+	Address            string   `yaml:"address"`              // Endpoint address, e.g. localhost:8080
+	CertFile           string   `yaml:"certfile"`             // A PEM eoncoded certificate file
+	KeyFile            string   `yaml:"keyfile"`              // A PEM encoded private key file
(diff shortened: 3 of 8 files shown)


ChristopherMacGown:
I've signed it

ChristopherMacGown:
I've signed it.

samples/server/go-server/api/server/api/api.go
```diff
@@ -97,8 +99,16 @@ func Run(config *Config, storage *server.Storager) {
 		log.Println("grpc server is configured without client certificate authentication")
 	}
 
+	if len(config.CORSAllowedOrigins) > 0 {
```
furuholm: I would prefer CORS to be disabled by default. It's easy enough to enable in the config.

vendor/vendor.json
```diff
@@ -369,5 +375,5 @@
 			"revisionTime": "2018-01-09T11:43:31Z"
 		}
 	],
-	"rootPath": "github.com/grafeas/grafeas"
```
furuholm: This doesn't look right.

ChristopherMacGown: Whoa, yes… thanks for catching that.

vendor/vendor.json
```diff
@@ -369,5 +375,5 @@
...
+	"rootPath": "github.com/glassbearinc/grafeas"
```
R2wenD2: Still not right, please use github.com/grafeas/grafeas

samples/server/go-server/api/server/api/api.go
ChristopherMacGown: Done.

vendor/vendor.json
ChristopherMacGown: You caught this in the middle of me making that edit. Sorry!

ChristopherMacGown:
I've signed it!

samples/server/go-server/api/server/api/api.go
```diff
@@ -29,16 +29,18 @@ import (
...
 type Config struct {
-	Address  string `yaml:"address"`  // Endpoint address, e.g. localhost:10000
-	CertFile string `yaml:"certfile"` // A PEM eoncoded certificate file
-	KeyFile  string `yaml:"keyfile"`  // A PEM encoded private key file
-	CAFile   string `yaml:"cafile"`   // A PEM eoncoded CA's certificate file
+	Address            string   `yaml:"address"`              // Endpoint address, e.g. localhost:10000
+	CertFile           string   `yaml:"certfile"`             // A PEM eoncoded certificate file
+	KeyFile            string   `yaml:"keyfile"`              // A PEM encoded private key file
+	CAFile             string   `yaml:"cafile"`               // A PEM eoncoded CA's certificate file
+	CORSAllowedOrigins []string `yaml:"cors_allowed_origins"` // Permitted CORS origins defaults to "*"
```
R2wenD2: Your comment indicates that this defaults to all origins? If that's the case, we definitely don't want that, prefer default to empty.

ChristopherMacGown: Yes, I'll update the docs and change that comment as it defaults closed.

docs/running_grafeas.md
```diff
@@ -156,4 +156,13 @@ func main() {
...
-```
\ No newline at end of file
+```
+
+## Enable CORS on the sample server.
+
+### Update config
+
+Add the following to your config file below the `api` key.
+    cors_allowed_origins:
```
furuholm: Use a fenced code block to keep consistent with the rest of the document https://help.github.com/articles/creating-and-highlighting-code-blocks/

config.yaml.sample
```diff
@@ -21,6 +21,9 @@ grafeas:
     cafile:
     keyfile:
     certfile:
+    cors_allowed_origins:
+       - "https://some.example.tld"
```
furuholm: I prefer the sample config to be the same as the default config. This means that the list of allowed origins should be empty by default. So either remove the example urls or move them to a comment.

docs/running_grafeas.md
ChristopherMacGown: I was trying to keep it consistent with the other [config example](https://github.com/grafeas/grafeas/blame/master/docs/running_grafeas.md#L83-L87) but didn't add the newline there. I'll add the newline to make it consistent with that other example.

furuholm: 👍 sounds good!

Issue #29, opened 2017-10-08
Implement filtering in reference implementation


R2wenD2:

hpejustin:
i think a go-restful based framework should be imported, https://github.com/hpejustin/depark this can be a good example.

hpejustin:
i think, we can add Label or Annotation properties to Note and Occurrence, in that case we can use filter to list them.

R2wenD2:
The intention here is to allow filtering on fields that already exist as a part of the occurrence or note. I don't see a case for additional labels or annotations.

The implementation will to parse a query string, reading key,values where keys are the fields of our resources.

R2wenD2:
We'll use one of the following filtering specs:

- https://github.com/google/cel-spec
- https://cloud.google.com/logging/docs/view/advanced_filters

Both of them require an accessible parser - I'm following up on that.

furuholm:
One aspect of this could be to go for a format that can be transformed into a database query, or will that be too restrictive?

R2wenD2:
We need a few things:
1) A parser that will parse and validate a filter expression from query string
2) A mapping of Keys to database fields (which I think is what you're referring to, @furuholm 

I'm working with the cel team to create part one of this, but once we have that part 2 can follow quickly.

We also need a real database, but that's a separate issue.

furuholm:
Added #84 to address the database

TristonianJones:
Hey there, I've implemented the filter parsing using the syntax of [advanced filters](https://cloud.google.com/logging/docs/view/advanced_filters) with the backing representation as [cel](https://github.com/google/cel-spec). Any preference on which directory the parser should go into?

R2wenD2:
@TristonianJones Awesome! Lets put it in grafeas/server-go/filtering.

TristonianJones:
Just curious, but any thoughts on whether this is something that you'd like to continue using? If so, the github.com/google/cel-go library can be used to evaluate ASTs generated by the filtering parser.

vtsao:
Yes we are still planning on using what you impl'd for our filtering. Also thanks for the code pointer. @aysylu FYI this might be something you will take on.

PR #105, opened 2018-02-14
Filter expression parser implementation.

TristonianJones:
The following change introduces an Antlr4 grammar that defines the filter syntax.
The grammar supports logical operations and sequences of restriction expressions,
with the ability to reference object fields and invoke functions supported by the
API as described in Grafeas Issue https://github.com/grafeas/grafeas/issues/29.

The syntax is equivalent to that supported by Google Cloud Logging [Advanced Filters](https://cloud.google.com/logging/docs/view/advanced-filters)
with one notable exception: it can accurately parse indexing operations on a variable
of the form:

    a[b]   // select field 'b' from identifier 'a'

Lack of support for the above syntax is a known limitation of Advanced Filters which
is not present here.

Parsed filters are output in the [Common Expression Language](https://github.com/google/cel-spec)
abstract syntax representation. A key difference between CEL and list filters is that
list filters may treat barewords in the expression as unquoted text. Within CEL, these
barewords would be flagged as identifiers. Whether the value is a string or identifier is
a contextual decision best left to the filter evaluation step.

Note, the CEL abstract syntax proto and Antlr4 Go runtime are vendored into this PR.


furuholm:
Great to see progress in this area!

For someone who has not been part of the discussion leading up to this PR: How is this supposed to be integrated with the existing code base? More specifically I'm thinking about the the storage layer implementations. Is there something in place that I missed (there are a lot of new files) or will this be handled separately?

R2wenD2:
@furuholm after we get this PR in, I'll work on a library that uses it to parse filters and map them to fields in the postgresql db.

TristonianJones:
Apologies on the file count @furuholm. The vendored Antlr4 Go runtime and generated source accounts for the bulk of the change. There are probably <1k lines of newly minted code under the filtering/ directory (despite what the contribution stats look like).

Happy to contribute. :)

furuholm:
@R2wenD2 Great!

@TristonianJones No worries! :)

Really looking forward to have great querying capabilities in Grafeas.

server-go/filtering/parser/parser_test.go
```diff
@@ -0,0 +1,121 @@
...
+	TestCaseDelimiter    = "\n\n"
+)
+
+type baseline struct {
+	expected *expr.ParsedExpr
+	errors   string
+	source   common.Source
+}
+
+func TestParse_Complex(t *testing.T) {
```
clrprod: This set of nested functions seems relatively complicated - is there any way we could just have a single test case which loads all the files and runs the tests inside a series of nested for loops?  (That way you can read the testcase without having to go look at function definitions).

server-go/filtering/common/source_test.go
```diff
@@ -0,0 +1,108 @@
...
+	str, found := source.Snippet(1)
+	assertTrue(t, found)
+	assertEquals(t, "hello, world", str)
+	str2, found := source.Snippet(2)
+	assertFalse(t, found)
+	assertEquals(t, "", str2)
+}
+
+// Helper assertions.
+func assertTrue(t *testing.T, actual bool) {
```
clrprod: Can we just inline these functions? I don't think they make the tests more readable. Alternatively can you call t.Helper() so the error messages include the actual line in the test function that causes the error?

TristonianJones: Yes, definitely. I'll inline them.

server-go/filtering/parser/parser.go
```diff
@@ -0,0 +1,417 @@
...
+	switch e.(type) {
+	case antlr.InputMisMatchException:
+		errorMsg = "Input mismatch"
+	case antlr.NoViableAltException:
+		errorMsg = "Unexpected token"
+	}
+	p.errors.ReportError(p.source, common.NewLocation(line, column), errorMsg)
+}
+
+func (p *parser) ReportAmbiguity(recognizer antlr.Parser, dfa *antlr.DFA,
```
R2wenD2: Can you add a function level comment here, including a description of why the method is intentionally left empty?

server-go/filtering/parser/parser.go
```diff
@@ -0,0 +1,417 @@
...
+	startIndex, stopIndex int, exact bool, ambigAlts *antlr.BitSet,
+	configs antlr.ATNConfigSet) {
+	// Intentional
+}
+
+func (p *parser) ReportAttemptingFullContext(recognizer antlr.Parser,
```
R2wenD2: Same here and below with function level comments.

R2wenD2:
Also, thanks for this, @TristonianJones

server-go/filtering/parser/parser_test.go
clrprod: If you need a message diffing library https://github.com/d4l3k/messagediff works.

re: IDE's and rerunning tests - I don't care strongly about whether there is a single table of tests, or you load each file in a separate test, I just just strongly encourage having tests be as simple as possible with minimal usage of helper functions and layers of abstraction.

The idea is that you want tests to be 'obviously correct' on inspection. Layers of function calls make it impossible to have this, and instead you have to go read each function and it's easier for bugs to slip in.

However I also approved this, so feel free to ignore my suggestion - this is mostly minor test style and I'm happy the tests exist at all :D

server-go/filtering/parser/parser.go
TristonianJones: Sure

server-go/filtering/parser/parser.go
TristonianJones: Of course, happy to add more comments.

server-go/filtering/parser/parser_test.go
TristonianJones: Thanks, I'm not sure I have time to restructure the tests just now, but will keep this in mind for the future.

TristonianJones:
Feel free to merge if you're happy with the responses to feedback.

PR #106, opened 2018-02-15
chore: use dep for dependency management

### LINKED ISSUE #62: Use golang/dep for managing the vendor tree
golang dep is the new/recommended way to handle deps.

JustinBeckwith:
This change removes the `vendor` directory from git, and sets up dependency management with [dep](https://golang.github.io/dep/).  

I am by no means a Makefile expert, so extra scrutiny there is advised :)

Resolves #62.


v1alpha1/proto/grafeas.swagger.json
```diff
@@ -1626,20 +1626,6 @@
       "description": "service Foo {\n      rpc Bar(google.protobuf.Empty) returns (google.protobuf.Empty);\n    }\n\nThe JSON representation for `Empty` is empty JSON object `{}`.",
       "title": "A generic empty message that you can re-use to avoid defining duplicated\nempty messages in your APIs. A typical example is to use it as the request\nor the response type of an API method. For instance:"
     },
-    "protobufFieldMask": {
```
vbatts: howcome this to be removed?

Makefile
```diff
@@ -5,29 +5,42 @@ CLEAN := *~
...
+	cd vendor/github.com/golang/protobuf/protoc-gen-go && go install .
+
+.install.googleapis:
+	mkdir -p vendor/github.com/googleapis && \
+		cd vendor/github.com/googleapis && \
+		git clone https://github.com/googleapis/googleapis
+	touch $@
 
 .install.grpc-gateway:
-	go get -u -v github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger && touch $@
```
vbatts: these can stay in a single command, and the `touch $@` ensures the target's timestamp so that it isn't run every time

furuholm:
To me it seems that removing the vendor directory is separate issue from switching to dep. Do we want to remove vendor? I don't have any experience with dep but my experience from using the other dependency management tools has been that keeping vendor versioned makes life way easier.

furuholm:
If dep can guarantee that you always get the same result I don’t mind removing vendor at all. It’s just been my experience that other packaging systems has not been able to do this reliably.

clrprod:
My understanding of
https://golang.github.io/dep/docs/ensure-mechanics.html#no-vendor-and-vendor-only

Is that dep ensure --vendor-only will populate your vendor directory with the correct dependencies as your lock file.

JustinBeckwith:
With vgo coming out, I'm not sure if this is the right direction anymore.  I'll let this one go and come back around :)

Issue #82, opened 2017-12-13
Consistent naming of Note and Occurrence kinds


furuholm:
Two of the Note kinds are postfixed "Type" (VulnerabilityType and BuildType) but not the rest of them and three of the Occurence kinds are postfixed "Details" (CustomDetails, VulnerabilityDetails and BuildDetails) but not the rest of them.

I think this should be made consistent while the API is considered unstable. I don't think Type is great, but I am fine with no postfix, Details or something else (I would however rather use Type than not fixing it).

R2wenD2:
This shouldn't be too difficult of a fix - and I'm supportive.
For notes, I'd be fine with just using the camel case name for the Type (here: https://github.com/Grafeas/Grafeas/blob/aae00b1f1853d8a253f98380c0232030f08d82eb/v1alpha1/proto/grafeas.proto#L642)

And for Occurrences, I like the details postfix because its giving details of a specific thing everywhere. 

But I'm also fine with using Details for both notes and occurrences.

I'm still working on grpc clients, but if I finish up there and you haven't sent an update I'll do this next.

furuholm:
Great! I have no strong opinions on where to use details or not so do what you think works best.

vtsao:
I'm in favor of consistent details suffix as well.

furuholm:
This seems to have been resolved in v1beta1 (see https://github.com/grafeas/grafeas/blob/master/v1beta1/proto/grafeas.proto). If you agree @R2wenD2 I will go ahead and close this issue.

vtsao:
Yup, we fixed this in v1beta1. Closing.

PR #110, opened 2018-03-07
Fix for Issue https://github.com/Grafeas/Grafeas/issues/82.

shrajfr12:
For occurrence, all fields in details have the Details suffix. For note however, neither making them all Details suffix nor making them all Type suffix seemed to match the name/usage pattern, so left them unchanged for now. Also protoc  3.5 on macOS would not compile as it was not finding the google/protobuf/*.proto (ubuntu was fine) so added the -I./include in Makefile so that we can build on macOS or ubuntu. Can change the note types also based on the feedback.


R2wenD2:
@shrajfr12 There should be changes to the generated files as well that you should check in.

shrajfr12:
Was not sure if generated files need to be checked in. Will do soon. Thx.

shrajfr12:
I see three files with new timestamps which are  grafeas.pb.go, grafeas.swagger.json and grafeas.pb.gw.go, but "git status" showed only the two files grafeas.pb.go and grafeas.swagger.json that could be committed so have included only those two.

R2wenD2:
@shrajfr12 Apologies for the delay here. Please update the branch and I'll merge.

Issue #100, opened 2018-01-25
Pagination support


furuholm:
We need pagination support to browse through the data (need arose after importing 18000 CVE:s from [Clair](https://github.com/coreos/clair)).

For postgres: I'm thinking we could use (encrypted) values of the id as page markers by default. This would work for insertion order listings at least. For other sort orders we would probably need to build separate indexes once they are supported land.

For memstore: I'm thinking we could use list index as page markers. This will not be bullet proof as you will get the wrong index if elements are deleted, but I think this is a decent trade off. The memstore should not be used in production.

The gRPC layer seems to have what is needed already.

R2wenD2:
Your suggestions here lgtm. Agree that we shouldn't spend a ton of time coming up with a solution for memstore.

furuholm:
Fixed by PR #115

PR #111, opened 2018-03-09
Pagination support (NOT READY TO MERGE)

furuholm:
This PR is a partial solution for #100. Only project pagination has been implemented. The reason for submitting a partial solution is to enable discussions before implementing pagination for all resource types.

Notes:
- I see a number of options for obfuscating page tokens:
  1. Don't
  2. Encode with e.g. base64
  3. Encrypt the tokens using a key that is only known to the server
  
  The memstore backend uses option 1in this PR (for simplicity).
  The postgres backend uses option 2 in this PR.
  Clair uses option 3.
  I started out with option 3 but figured I would keep it simple until I got some feedback. What is the preferred option   for the postgres backend? Is there any best practice at google? 

- I added a ListOptions struct to keep the number of parameters down. Not sure if this was an improvement or not. Feedback welcome.

- We might have to add new db indexes if the upcomming filtering functionality will support ordering.

### DIFF
--- a/samples/server/go-server/api/server/storage/memstore.go
+++ b/samples/server/go-server/api/server/storage/memstore.go
@@ -85,11 +85,12 @@ func (m *memStore) GetProject(pID string) (*pb.Project, error) {
 	return project, nil
 }
 
-// ListProjects returns the project id for all projects from the mem store
-func (m *memStore) ListProjects(options *server.ListOptions) ([]*pb.Project, string, error) {
+// ListProjects returns up to pageSize number of projects beginning at pageToken (or from
+// start if pageToken is the emtpy string).
+func (m *memStore) ListProjects(filter string, pageSize int, pageToken string) ([]*pb.Project, string, error) {
 	m.RLock()
 	defer m.RUnlock()
-	startPos, endPos, err := calcRange(options, len(m.projects))
+	startPos, endPos, err := calcRange(pageSize, pageToken, len(m.projects))
 	if err != nil {
 		return nil, "", err
 	}
@@ -316,17 +317,16 @@ func (m *memStore) ListOperations(pID, filters string) ([]*opspb.Operation, erro
 	return ops, nil
 }
 
-// Calculates start and end positions in a range defined by provided ListOptions and upperLimit
-func calcRange(options *server.ListOptions, upperLimit int) (startPos, endPos int, err error) {
-	pageSize := int(options.PageSize)
-	if options.PageToken != "" {
-		startPos, err = strconv.Atoi(options.PageToken)
+// Calculates start and end positions given the provided constraints
+func calcRange(pageSize int, pageToken string, upperLimit int) (startPos, endPos int, err error) {
+	if pageToken != "" {
+		startPos, err = strconv.Atoi(pageToken)
 		if err != nil {
-			return 0, 0, status.Error(codes.InvalidArgument, fmt.Sprintf("PageToken invalid: %s", options.PageToken))
+			return 0, 0, status.Error(codes.InvalidArgument, fmt.Sprintf("PageToken invalid: %s", pageToken))
 		}
 	}
 	if startPos < 0 || startPos >= upperLimit {
-		return 0, 0, status.Error(codes.OutOfRange, fmt.Sprintf("PageToken out of range: %s", options.PageToken))
+		return 0, 0, status.Error(codes.OutOfRange, fmt.Sprintf("PageToken out of range: %s", pageToken))
 	}
 	return startPos, min(startPos+pageSize, upperLimit), nil
 }
--- a/samples/server/go-server/api/server/storage/memstore.go
+++ b/samples/server/go-server/api/server/storage/memstore.go
@@ -94,7 +94,7 @@ func (m *memStore) ListProjects(filter string, pageSize int, pageToken string) (
 	sort.Slice(projects, func(i, j int) bool {
 		return projects[i].Name < projects[j].Name
 	})
-	startPos, endPos, err := calcRange(pageSize, pageToken, len(m.projects))
+	startPos, endPos, err := calcRange(pageSize, pageToken, len(projects))
 	if err != nil {
 		return nil, "", err
 	}
@@ -159,7 +159,14 @@ func (m *memStore) ListOccurrences(pID, filters string, pageSize int, pageToken 
 			os = append(os, o)
 		}
 	}
-	return os, "", nil
+	sort.Slice(os, func(i, j int) bool {
+		return os[i].Name < os[j].Name
+	})
+	startPos, endPos, err := calcRange(pageSize, pageToken, len(os))
+	if err != nil {
+		return nil, "", err
+	}
+	return os[startPos:endPos], strconv.Itoa(endPos), nil
 }
 
 // CreateNote adds the specified note to the mem store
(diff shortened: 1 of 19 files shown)


server-go/storage.go
```diff
@@ -19,6 +19,16 @@ import (
 	opspb "google.golang.org/genproto/googleapis/longrunning"
 )
 
+type ListOptions struct {
```
R2wenD2: @vtsao Do you have thoughts on this struct? I have no issues with it.

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -42,53 +43,62 @@ func NewMemStore() server.Storager {
...
 	}
-	delete(m.projects, pID)
+	copy(m.projects[index:], m.projects[index+1:])
+	m.projects[len(m.projects)-1] = nil
+	m.projects = m.projects[:len(m.projects)-1]
 	return nil
 }
 
 // GetProject returns the project with the given pID from the mem store
 func (m *memStore) GetProject(pID string) (*pb.Project, error) {
```
R2wenD2: Perhaps we should keep the map for efficient gets and then just use the sorted keys of the map for the List method? What do you think? 

This isn't a huge deal since its in memory and the size of things shouldn't be too large.

samples/server/go-server/api/server/storage/memstore.go
```diff
@@ -42,53 +43,62 @@ func NewMemStore() server.Storager {
 		occurrencesByID: map[string]*pb.Occurrence{},
 		notesByID:       map[string]*pb.Note{},
 		opsByID:         map[string]*opspb.Operation{},
```
R2wenD2: What are your plans for pagination on these? We can get all of the values for a project and sort them, I guess.

jzelinskie:
You should probably use [fernet](https://godoc.org/github.com/fernet/fernet-go) rather than base64 to avoid leaking primary keys from the database. This is what clair uses.

R2wenD2:
Sorry - I wasn't clear, I meant an encrypted base64 encoded string. fernet also looks fine.

Agree that we need to support migrations, we'll eventually get to that point :)

samples/server/go-server/api/server/storage/memstore.go
furuholm: Agree. Switching to lists turned out to be more intrusive than I first thought so going back to a map is better for several reasons.

samples/server/go-server/api/server/storage/memstore.go
furuholm: Will add pagination for them as well. Just wanted the projects to be done first. And yes, if we sort them we should be fine.

furuholm:
Ok. I'll add page token encryption.

server-go/storage.go
vtsao: I would just inline these fields as args to the method, it makes it easier to read and I don't think we expect this to grow anytime soon.

R2wenD2:
@furuholm is this ready for another look?

furuholm:
@R2wenD2 I haven't got around to it. I will try to get some time to do it during this week. Do you have other work that is being held up by this?

R2wenD2:
no blockers on my part. just wanted to make sure you weren't waiting on me for a review :)

server-go/storage.go
furuholm: Fixed in 724e2e3

samples/server/go-server/api/server/storage/memstore.go
furuholm: OMG. So much ugly code was introduced by this change. What was I thinking?!? Good thing you reviewed this.😄 

363992e reverts back to storing projects in a map.

furuholm:
Tokens are now encrypted using fernet-go.

Will keep this PR open until we are satisfied with project pagination. Will create a new PR that adds pagination for the other entity types.

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -126,22 +130,36 @@ func (pg *pgSQLStore) GetProject(pID string) (*pb.Project, error) {
 	return &pb.Project{Name: pName}, nil
 }
 
-// ListProjects returns the project id for all projects from the store
-func (pg *pgSQLStore) ListProjects(filters string) ([]*pb.Project, error) {
-	rows, err := pg.DB.Query(listProjects)
+var paginationKey = "XxoPtCUzrUv4JV5dS+yQ+MdW7yLEJnRMwigVY/bpgtQ="
```
vtsao: Is this meant to be here, shouldn't this be read from the config?

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -126,22 +130,36 @@ func (pg *pgSQLStore) GetProject(pID string) (*pb.Project, error) {
...
+	var rows *sql.Rows
+	id, err := decryptInt64(pageToken, paginationKey)
+	if err == nil {
+		rows, err = pg.DB.Query(listProjectsFromPage, pageSize, id)
+	} else {
+		rows, err = pg.DB.Query(listProjects, pageSize)
+	}
 	if err != nil {
-		return nil, status.Error(codes.Unknown, fmt.Sprintf("Failed to list Projects from database"))
+		return nil, "", status.Error(codes.Unknown, "Failed to list Projects from database")
```
vtsao: codes.Internal is probably better here and below

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -472,3 +490,23 @@ func (pg *pgSQLStore) ListOperations(pID, filters string) ([]*opspb.Operation, e
 	}
 	return ops, nil
 }
+
+// Encrypt int64 using provided key
+func encryptInt64(v int64, key string) (string, error) {
+	k, _ := fernet.DecodeKey(key)
```
vtsao: You should check the error here, same for decrypt.

samples/server/go-server/api/server/storage/pgsqlstore.go
```diff
@@ -472,3 +490,23 @@ func (pg *pgSQLStore) ListOperations(pID, filters string) ([]*opspb.Operation, e
...
+	bytes, err := fernet.EncryptAndSign([]byte(strconv.FormatInt(v, 10)), k)
+	if err != nil {
+		return "", err
+	}
+	return string(bytes), nil
```
vtsao: we should encode this in base64

samples/server/go-server/api/server/storage/pgsqlstore.go
furuholm: You are right. Forgot to remove this. Fixed in 5108566.

samples/server/go-server/api/server/storage/pgsqlstore.go
furuholm: Agree. Fixed in 2c060cf

samples/server/go-server/api/server/storage/pgsqlstore.go
furuholm: The token returned from fernet-go is already base64 encoded.

furuholm:
@vtsao: I've pushed fixes that addresses your comments, but I didn't rebase the changes so there are a a bunch of other things that has been added to this PR that really should be put in a new PR. You can ignore them for now if you want to.

furuholm:
I've created a new PR #115 where I have squashed and rebased the commits.

(I've pushed all of the commits that was squashed into #115 to this PR if anyone prefers reviewing individual changes.)

PR #115, opened 2018-03-26
Pagination support for projects, notes, occurrences and operations

furuholm:
Squashed and rebased version of #111

Question to reviewers: I've defaulted the page size to 100 for requests that does not specify page size. Sounds reasonable?


vtsao:
> Question to reviewers: I've defaulted the page size to 100 for requests that does not specify page size. Sounds reasonable?

That's fine for now. Eventually we should make page size a required param when listing things, since entities like occurrences can grow to be very large.

vtsao:
Thanks for all your work on this!

PR #125, opened 2018-04-07
Modified code to ignore postgres settings when storage type is memstore

arunbsar:
Modified code to ignore postgres settings when storage type is memstore, this allows removing postgres setting from config file when not used or will generate runtime error "invalid memory address or nil pointer dereference"

### DIFF
--- a/samples/server/go-server/api/server/config/config.go
+++ b/samples/server/go-server/api/server/config/config.go
@@ -67,6 +67,11 @@ func LoadConfig(fileName string) (*config, error) {
 		return nil, err
 	}
 	config := configFile.Grafeas
+
+	if config.StorageType == "memstore" {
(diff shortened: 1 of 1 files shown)


samples/server/go-server/api/server/config/config.go
```diff
@@ -67,6 +67,11 @@ func LoadConfig(fileName string) (*config, error) {
 		return nil, err
 	}
 	config := configFile.Grafeas
+
+	if config.StorageType == "memstore" {
```
furuholm: Great addition! I'd prefer the if clause to test for postgres (instead of memstore) and wrap the key creation code below. While it doesn't matter much today this will be better when other changes are made (like adding another storage type).

PR #129, opened 2018-05-02
rename note_type to just "type" since it's already under the note

vtsao:
Rename note_type to just "type" since it's already under the note.


vtsao:
Looks like protoc gen library was updated, need to update vendor deps for "make" to be happy. PR #130 is out for that.

PR #151, opened 2018-06-05
update GetVulnerabilityOccurrencesSummary to show fixable/total by

R2wenD2:
severity

### DIFF
--- a/v1beta1/proto/grafeas.proto
+++ b/v1beta1/proto/grafeas.proto
@@ -566,24 +566,18 @@ message GetVulnerabilityOccurrencesSummaryRequest {
 
 // A summary of how many vulnerability occurrences there are per severity type.
 message VulnerabilityOccurrencesSummary {
-  // A map of how many occurrences were found for each severity.
-  repeated SeverityCount counts = 1;
   // A listing by resource of the number of fixable and total vulnerabilities.
-  repeated FixableTotalByDigest fixable_counts = 2;
-  // The number of occurrences for a specific severity.
-  message SeverityCount {
-    // The severity of the occurrences.
-    grafeas.v1beta1.vulnerability.Severity severity = 1;
-    // The number of occurrences with the severity.
-    int64 count = 2;
-  }
-  // Per resource counts of fixable and total vulnerabilites.
+  repeated FixableTotalByDigest fixable_counts = 1;
+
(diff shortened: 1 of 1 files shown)


v1beta1/proto/grafeas.proto
```diff
@@ -566,24 +566,18 @@ message GetVulnerabilityOccurrencesSummaryRequest {
 
 // A summary of how many vulnerability occurrences there are per severity type.
```
vtsao: per resource and severity type.

v1beta1/proto/grafeas.proto
```diff
@@ -566,24 +566,18 @@ message GetVulnerabilityOccurrencesSummaryRequest {
...
-  repeated FixableTotalByDigest fixable_counts = 2;
-  // The number of occurrences for a specific severity.
-  message SeverityCount {
-    // The severity of the occurrences.
-    grafeas.v1beta1.vulnerability.Severity severity = 1;
-    // The number of occurrences with the severity.
-    int64 count = 2;
-  }
-  // Per resource counts of fixable and total vulnerabilites.
+  repeated FixableTotalByDigest fixable_counts = 1;
```
vtsao: just count now? since we've combined them

vtsao: or "counts" rather

v1beta1/proto/grafeas.proto
```diff
@@ -566,24 +566,18 @@ message GetVulnerabilityOccurrencesSummaryRequest {
...
+
+  // Per resource and severity counts of fixable and total vulnerabilites.
   message FixableTotalByDigest {
```
vtsao: Probably should just rename this to "message Counts" in case we decide to change its breakdown later?

R2wenD2: I'd prefer to leave it more specific. If we change it later we'd need to version anyway.

v1beta1/proto/grafeas.proto
R2wenD2: done

v1beta1/proto/grafeas.proto
R2wenD2: done

vtsao:
Did you commit? I don't see any changes.

R2wenD2:
committed now, I need to fix my tooling.

PR #152, opened 2018-06-07
removing operation_name from notes and occurrences

vtsao:
We don't have a good use for operations right now.


vtsao:
I'll submit another PR for that.

Issue #157, opened 2018-06-13
occurrence IDs should be generated by the API


vtsao:
In our reference impl of the v1alpha1 Grafeas API, we currently require the occurrence name to be set in CreateOccurrence:
https://github.com/grafeas/grafeas/blob/master/samples/server/go-server/api/server/v1alpha1/impl.go#L101

Occurrence names are output only and their IDs should be generated by the API:
https://github.com/grafeas/grafeas/blob/master/v1alpha1/proto/grafeas.proto#L774

Note this is unlike notes, which you need to set the ID for when creating.

This should be fixed in the v1alpha1 ref impl, or it can wait until we update the ref impl to v1beta1.

vtsao:
FYI @shrajfr12

vtsao:
Dan assigning to you for now. This will go hand in hand with the Go client lib sample revamp.

DanSanche:
I'll take a look. Do we want the server to reject any occurrences with assigned names (to give the user feedback), or should it just silently assign and use its own names?

DanSanche:
I'm addressing this here: https://github.com/grafeas/grafeas/pull/160

aysylu:
#160 appears to have addressed the issue. Is there anything else?

vtsao:
#160 did this in the v1alpha1 API layer. This needs to be done in the storage layer, so there is still work here.

aysylu:
I'll take this for the server implementation that uses v1beta1 API.

PR #160, opened 2018-06-14
Server-Side Occurrence Names

DanSanche:
Addressing the following issue: https://github.com/grafeas/grafeas/issues/157

Now, the server will automatically assign an name to new occurrences, and reject any that try to set an explicit one

I made use of a UUID library to generate the names, so they will look something like "projects/best-vuln-scanner/occurrences/fadad2f2-6899-446d-9428-8876c75c7962"


DanSanche:
It looks like some tests are failing because it assumes it can set and retrieve specific occurrence names, so I'll have to fix the tests.



Also, right now I'm rejecting any occurrences that specify a name, so the user can get some feedback. This also breaks some tests. Let me know if you'd prefer it if the server just changed the name silently

mikecico:
It might be useful to have the name have some user-friendly part, e.g., the project name as a prefix, and/or allow some customizable prefix or suffix.

vtsao:
@DanSanche server should silently ignore any output only fields users set:
https://cloud.google.com/apis/design/design_patterns#output_fields

@mikecico that's an interesting proposal - can you outline your use case for this?

DanSanche:
@vtsao I've addressed your comments, and the tests are now passing

mikecico:
@vtsao we have a use case where we might be storing occurrences from multiple scanners within the same project (the other option of course is to have separate projects for each one, but we're trying to consolidate scanning steps within a pipeline job/step).   Since prior to this the naming field was set by the client, we were thinking of creating a shallow hierarchy through naming conventions, which would also make the occurrences more readable, and allow us to be able to categorize occurrences within the project that we can create searches on, etc.

I think perhaps allowing clients influence on the naming field might not be necessary, and its probably better to leave it under the control of Grafeas if it is intended to be a generated ID.  But perhaps if there were something akin to an optional "display-name" field, or something under client control, that might be useful.  Unless there's something we can already leverage for this purpose?

raj-inamdar:
To second @mikecico 's comment above, it will be useful to be able to add labels/tags to occurrences so they can be filtered/queried effectively.

R2wenD2:
@mikecico and @raj-inamdar If you're storing occurrences from two scanners, the correct thing to do here would be to use separate not projects for each scanner, and then you could differentiate the occurrences by note name. Is there a reason that wouldn't work?

raj-inamdar:
@R2wenD2 In general, we expect to create separate notes projects for each scanner and use the note-name in the occurrences to differentiate, just as you suggest.

However, we think that there may be situations where multiple scanners may have some overlap and may identify the same underlying condition. Instead of duplicating notes in respective projects for such scanners, we may want to create a "shared" notes project. In that case, the note-name in occurrences may not be sufficient to differentiate across different scanners.

Besides, we think that a user-costomizable labels/tags field may be useful in general to add flexibility for filtering, querying, etc.

vtsao:
@mikecico @raj-inamdar thanks for the use case description. Can either of you open an issue so we can track/discuss it there? Thanks.

samples/server/go-server/api/server/v1alpha1/impl.go
```diff
@@ -96,42 +97,45 @@ func (g *Grafeas) CreateOccurrence(ctx context.Context, req *pb.CreateOccurrence
...
 		}
-		if _, err = g.S.GetOperation(pID, oID); err != nil {
-			log.Printf("Operation:%v for Occurrence: %v not found", oID, o.Name)
-			return nil, status.Error(codes.NotFound, fmt.Sprintf("Operation:%v for Occurrence: %v not found", oID, o.Name))
+		if _, err = g.S.GetOperation(opID, oID); err != nil {
+			log.Printf("Operation:%v for Project: %v not found", oID, opID)
+			return nil, status.Errorf(codes.NotFound, "operation: %v for project: %v not found", oID, opID)
 		}
 	}
+	randID, err := uuid.NewRandom()
```
vtsao: You can do this in a follow-up PR, but we should move the UUID generation to the storage layer, since it's a storage concept. Depending on the DB being used, some people may choose to let the DB create this ID for instance.

raj-inamdar:
@vtsao Entered https://github.com/grafeas/grafeas/issues/163 to track our request. Thanks.

sandeeps2:
I have a comment, hope it is not too late. In general it might be useful to have a human readable name. We can keep the original name field which is provided by the user in the CreateOccurrenceRequest and introduce a new ID field which is generated by the Server and returned to user after a successful create. It will keep the API same but existing queries will break, we can address it with a new method to ListOccurrencesByName.

vtsao:
@raj-inamdar thanks. @sandeeps2 can you add your comment on #163.

sandeeps2:
@vtsao While a general purpose labeling facility will be useful, it is good to have Name as a first class field which can be indexed and easily searched for an Occurrence. My comment was more for the in-flight change for this Issue to retain the Name field of the Occurrence as a user-supplied value and define a new field for the generated ID. I will add my related thoughts on #163 as well. 
Thanks.

vtsao:
Is this ready to be merged now Dan?

DanSanche:
Yes, I believe it should be. It looks like I don't have merge permissions for the repository

PR #184, opened 2018-08-11
marking some required fields for deployment msgs

vtsao:


v1beta1/proto/deployment.proto
```diff
@@ -25,7 +25,7 @@ import "google/protobuf/timestamp.proto";
 
 // An artifact that can be deployed in some runtime.
 message Deployable {
-  // Resource URI for the artifact being deployed.
+  // Required. Resource URI for the artifact being deployed.
```
vtsao: Is this also immutable? Can it be updated?

PR #197, opened 2018-08-20
vuln details should be required

vtsao:
Otherwise the vuln is not very useful.


vtsao:
WANT_LGTM=ANY

Katee:
I don't think that is required. A CVE number can be assigned with a description before it is in the os specific feeds.

vtsao:
Yeah that's true. Closing this.

PR #199, opened 2018-08-22
documenting contraints for batch create and list methods

vtsao:


v1beta1/proto/grafeas.proto
```diff
@@ -375,10 +376,14 @@ message ListNotesRequest {
   // The name of the project to list notes for in the form of
   // `projects/[PROJECT_ID]`.
   string parent = 1;
+
   // The filter expression.
   string filter = 2;
-  // Number of notes to return in the list.
+
+  // Must be positive. Max allowed page size is 1000. Number of notes to return
```
Katee: [nit] I think what it is goes first, then the default value and limitations.

vtsao: Done.

PR #205, opened 2018-08-28
Add embedded db to storage engines

liron-l:
Add embedded-db storage engine based on
[boltdb](https://github.com/boltdb/bolt).
Embedding the storage directly in grafeas provides deployment simplicity
and persistency (assuming a persistent volume is used).

### DIFF
--- a/samples/server/go-server/api/server/storage/embeddedstore.go
+++ b/samples/server/go-server/api/server/storage/embeddedstore.go
@@ -1,0 +1,415 @@
+// Copyright 2017 The Grafeas Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/boltdb/bolt"
+	"github.com/golang/protobuf/proto"
+	"github.com/grafeas/grafeas/samples/server/go-server/api/server/name"
+	"github.com/grafeas/grafeas/server-go"
+	pb "github.com/grafeas/grafeas/v1alpha1/proto"
+	opspb "google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+const (
+	bucketOccurrences = "occurrences"
+	bucketProjects    = "projects"
+	bucketNotes       = "notes"
+	bucketOperations  = "operations"
+)
+
+var (
+	errKeyExists = fmt.Errorf("key exists")
+	errNoKey     = fmt.Errorf("key missing")
+)
+
+type EmbeddedStoreConfig struct {
+	Path string `yaml:"path"` // Path is the folder path to storage files
+}
+
+// embeddedStore is a storage solution for Grafeas based on boltdb
+type embeddedStore struct {
+	db *bolt.DB
+}
+
+// NewEmbeddedStore creates a embeddedS store with initialized filesystem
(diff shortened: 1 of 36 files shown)


furuholm:
Nice! Could you elaborate on the use case for this storage type? While adding a new storage type adds value, it also adds future maintenance cost. One question in particular that needs to be answered before adding any new storage backend is what type of support it should have regarding database migrations as changes to the data model are introduced.

liron-l:
Hi @furuholm, thanks for your reply.

1. The main reasoning for this change is to simplify cluster deployment. That is, we would like to enable running a standalone production-ready image of Grafeas without external dependencies on storage (which require authz/authn and additional connectivity configuration). 
Boltdb is a robust embedded storage which is used in other cloud native projects (e.g., InfluxDB  and docker swarm). 
I considered using sqlite, but the C embedding makes the build process slower, and i prefer a native go solution.

2. Re schema migration, the current implementation store each schema in a dedicated bucket (collection)  where the value is the serialized value. Thus, we can do data migration of a bucket by iterating a collections key/values and modifying/adjusting the data. 

Let me know if this answers your questions.

furuholm:
Thanks for the answers! I'm ok with this. Just wanted to highlight some things that might be good to for @vtsao to consider when doing the cost/benefit analysis of adding another storage backend.

Another thing that might be worth having in mind is how this integrates with #29.

vtsao:
Hi @liron-l, thanks for your PR. We're currently discussing the maintenance implications of adding another storage layer, as @furuholm mentioned as well. We will get back to you.

liron-l:
Thanks @furuholm and @vtsao.
If you're concerned about maintenance, I think an embedded engine might replace the existing in-memory engine. They both provide the same prototyping benefits while the embedded solution also provides persistenty. 
Also, if we add support for exposing a unix socket, we get a default secure and standalone solution.

furuholm:
Replacing the memory store with a persistent one could make sense but it will be crucial to establish and communicate the level of migration support this solution provides. This is not an issue for the memory store as it (obviously) does not provide any support for migrations.

furuholm:
But you still have to implement the migrations which you never would have to do for the memory store. This means two migration systems (Postgres and BoltDB) have to be maintain if the BoltDB should have the same level of support as the Postgres store.

_(BTW: I'm not trying to beat on this PR. Just want all the information to be on the table)_

liron-l:
@furuholm yes that's correct. 
The benefit is that together with unix socket, you can have a secure local standalone solution without external dependencies.
I can achieve the same final results by creating a complex image, but trying to avoid this this if possible.

furuholm:
If adding the storage seems like to much of a commitment an alternative could be to do what CoreOS Clair does. They allow components to register themselves dynamically using a pattern described [here](https://github.com/coreos/clair/blob/master/Documentation/drivers-and-data-sources.md). @jzelinskie mentioned that they had used this mechanism to allow users to extend Clair without having to take over the ownership of the code.

This pattern would have to be added to Grafeas as well. Once that is done it is possible to add another storage implementation by dropping the code into a package that is imported by main.go. The downside to this solution is that it requires the source code to be patched. OTOH it allows adding new implementations without having to get approval from the maintainers.

liron-l:
Thanks @furuholm. I think the use-case in clair is adding a storage driver to a private fork (so you can use it in your own implementation) and not to the mainline code, where 'open-sourcing' the driver is equivalent to submitting the engine in the source code (like we do here). 
I do think that adding an embedded storage to this project can benefit other users, which is the reason i suggested to included it as a first class citizen in the project.

liron-l:
Thanks @furuholm and @vtsao for the feedback. 
Do we need additional code-reviewers for this change?

samples/server/go-server/api/server/storage/embeddedstore.go
```diff
@@ -0,0 +1,415 @@
...
+		log.Fatal(err)
+	}
+	return &embeddedStore{db: db}
+}
+
+// CreateProject adds the specified project to the embedded store
+func (m *embeddedStore) CreateProject(pID string) error {
+	err := m.update(bucketProjects, pID, true, &pb.Project{Name: name.FormatProject(pID)})
+	if err == errKeyExists {
+		return status.Error(codes.AlreadyExists, fmt.Sprintf("Project with name %q already exists", pID))
```
vtsao: You can use status.Errorf for all of these.

liron-l: Thanks @vtsao. I've replaced all usages of `status.Error` with `status.Errorf` for `embeededStore` `pgSQLStore` and `memStore`.

PR #209, opened 2018-08-28
Adding Go package option to protos and fixing make file some more.

vtsao:


vtsao:
Adding Go package option so protoc generates Go proto bindings with the correct imports. Also adding the new targets to `make build` and phonies.

vtsao:
We generate gRPC Go proto binding reverse proxies for the services, so add `mv` rules to move them to the correct folder after generation.

vtsao:
Gotta bootstrap the v1beta1 proto folders to exist first. Removing v1beta1 targets from build make target for now.

PR #216, opened 2018-08-30
Support running grafeas server with unix socket

liron-l:
Enables running a secure standalone grafeas deployment without TLS

### DIFF
--- a/samples/server/go-server/api/server/api/api.go
+++ b/samples/server/go-server/api/server/api/api.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/cockroachdb/cmux"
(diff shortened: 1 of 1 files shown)


samples/server/go-server/api/server/api/api.go
```diff
@@ -45,11 +46,18 @@ type Config struct {
 
 // Run initializes grpc and grpc gateway api services on the same address
 func Run(config *Config, storage *server.Storager) {
-	l, err := net.Listen("tcp", config.Address)
+	network, address := "tcp", config.Address
+	if strings.Contains(config.Address, "unix://") || strings.Contains(config.Address, "/") {
```
vtsao: Maybe strings.HasPrefix instead, since it should always start with "unix://" right?

Also, not sure we want to assume unix socket if we find a "/" in general, since you could have something like localhost:8080/api as a TCP endpoint.

liron-l: Makes sense @vtsao. Let me know if you think we should completely removes the `/` prefix.

liron-l:
Thanks @vtsao! updated the review according to comments

samples/server/go-server/api/server/api/api.go
vtsao: I think we should remove the or check, so remove this:
|| strings.HasPrefix(config.Address, "/")

liron-l: Sure @vtsao makes sense. I removed the second check and updated the code.

PR #218, opened 2018-09-17
Return empty page token for last response

liron-l:
According to protobuf specification, the next token should be empty for
the last.
```go
	// The next pagination token in the list response. It should be used as
	// page_token for the following request. An empty value means no more result.
	NextPageToken        string   `protobuf:"bytes,2,opt,name=next_page_token,json=nextPageToken,proto3" json:"next_page_token,omitempty"`
```

### DIFF
--- a/samples/server/go-server/api/server/storage/storager_test.go
+++ b/samples/server/go-server/api/server/storage/storager_test.go
@@ -444,9 +444,12 @@ func doTestStorager(t *testing.T, createStore func(t *testing.T) (server.Storage
 			wantProjectNames = append(wantProjectNames, name.FormatProject(pID))
 		}
 		filter := "filters_are_yet_to_be_implemented"
-		gotProjects, _, err := s.ListProjects(filter, 100, "")
+		gotProjects, pageToken, err := s.ListProjects(filter, 100, "")
 		if err != nil {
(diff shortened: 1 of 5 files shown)


samples/server/go-server/api/server/storage/storager_test.go
```diff
@@ -444,10 +444,13 @@ func doTestStorager(t *testing.T, createStore func(t *testing.T) (server.Storage
 			wantProjectNames = append(wantProjectNames, name.FormatProject(pID))
 		}
 		filter := "filters_are_yet_to_be_implemented"
-		gotProjects, _, err := s.ListProjects(filter, 100, "")
+		gotProjects, pageToken, err := s.ListProjects(filter, 100, "")
 		if err != nil {
 			t.Fatalf("ListProjects got %v want success", err)
 		}
+		if pageToken != "" {
+			t.Errorf("ListProjects should return empty page token, got %s", pageToken)
```
vtsao: Generally we do got before want for test errors. E.g.,
Got %s, want empty page token.

liron-l: Thanks @vtsao fixed.

vtsao:
Thanks for the PR @liron-l

PR #220, opened 2018-09-18
Migrate sample server from v1alpha1 to v1beta1 - schema

liron-l:
Partial migration from v1alpha1 to v1beta1 -
The only difference between versions is that v1beta does not have the
`operation` notation so i completely removed it from the server
implementation.

Remaining work:
1. Implement the missing storage functions: `BatchCreateNotes`,
`BatchCreateOccurrences` and `GetVulnerabilityOccurrencesSummary`
2. Delete unused storage function (operation related)
3. Move `impl.go` from folder `v1alpha1` to folder `v1beta1`

I can do the reaming work in this commit, or do it in a separate commit


vtsao:
This PR looks fine, but I want to hold off on merging it. As mentioned in the issue thread, let's discuss at the next Grafeas sync.

Specifically we are planning to remove the [samples](https://github.com/grafeas/grafeas/tree/master/samples) folder and use the [go](https://github.com/grafeas/grafeas/tree/master/go) folder instead as we have opensourced our internal implementation of the v1beta1 API layer. So the v1alpha1 API layer you've updated in this PR is going to be removed.

The [storage](https://github.com/grafeas/grafeas/tree/master/samples/server/go-server/api/server/storage) code is going to be updated to work with the new [v1beta1 interfaces](https://github.com/grafeas/grafeas/tree/master/go/v1beta1/api).

But let's discuss more at the sync. Thanks for your contributions so far!

PR #224, opened 2018-09-27
Allows deletion of occurrences which have an empty note id

tylerdatgoogle:
These occurrences cannot pass the authorization check for note attach permissions, so currently cannot be deleted. But if no note is provided, they should be allowed to be deleted anyway.

### DIFF
--- a/go/name/name.go
+++ b/go/name/name.go
@@ -41,6 +41,9 @@ func ParseProject(name string) (string, error) {
 
 // ParseNote parses the project ID and note ID from a note resource name.
 func ParseNote(name string) (string, string, error) {
+	if name == "" {
+		return "", "", nil
+	}
 	params := strings.Split(name, "/")
 	if len(params) != 4 {
 		return "", "", errors.Newf(codes.InvalidArgument, "name must be in the form 'projects/[PROJECT_ID]/notes/[NOTE_ID]', got %q", name)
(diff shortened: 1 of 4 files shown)


go/v1beta1/api/occurrence.go
```diff
@@ -245,8 +245,10 @@ func (g *API) DeleteOccurrence(ctx context.Context, req *gpb.DeleteOccurrenceReq
 	if err != nil {
 		return err
 	}
-	if err := g.Auth.CheckAccessAndProject(ctx, notePID, nID, NotesAttachOccurrence); err != nil {
-		return err
+	if notePID != "" && nID != "" {
```
vtsao: I'd prefer it if we checked if o.NoteName were empty instead of changing the parser.

go/name/name.go
```diff
@@ -41,6 +41,9 @@ func ParseProject(name string) (string, error) {
 
 // ParseNote parses the project ID and note ID from a note resource name.
 func ParseNote(name string) (string, string, error) {
+	if name == "" {
```
vtsao: As mentioned below, I think we can just check if o.NoteName is empty, I don't think we want to change the parser.

go/v1beta1/api/occurrence.go
tylerdatgoogle: done

go/name/name.go
tylerdatgoogle: done

PR #234, opened 2018-10-29
docs: fix markdown rendering

sublimino:
This is also fixable by changing `<Checksum>` to `checksum` and avoiding markdown code blocks, although code blocks are clearer IMHO.

Also fixed in Grafeas docs https://github.com/grafeas/grafeas.github.io/pull/19

---

FWIW I fixed this in March, but suspect that the repo docs were pasted into the website, so have now updated both https://github.com/grafeas/grafeas.github.io/pull/5

### DIFF
--- a/docs/grafeas_concepts.md
+++ b/docs/grafeas_concepts.md
@@ -48,8 +48,8 @@ timestamp. Where possible, use content-addressable resource URLs.
 The following table provides examples of possible resource URLs for several
 component types:
 
-Component Type | Identifier                               | Example
--------------- | ---------------------------------------- | -------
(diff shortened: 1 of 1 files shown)


docs/grafeas_concepts.md
```diff
@@ -50,14 +50,14 @@ component types:
 
 Component Type | Identifier                               | Example
 -------------- | ---------------------------------------- | -------
```
vtsao: Can you fix the alignment of the header.

vtsao:
Thanks for the fix @sublimino !

PR #239, opened 2019-01-10
Run circlci on both PR's as well as nightly to detect failures

clrprod:


.circleci/config.yml
```diff
@@ -49,3 +49,18 @@ jobs:
...
       - run: make build; make test
+workflows:
+  version: 2
+  commit:
+    jobs:
+      - build
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
```
aysylu: nit: comment in human readable form on when this will run, please

PR #243, opened 2019-01-11
Add Windows vulnerabilities to the vulnerability detail format.

clrprod:
Note that we give the Windows vulnerabilities their own format due
to different patch semantics. This prevents existing systems from
incorrectly handling them, and allows linux only users to ignore
them.


vtsao:
I'll let @aysylu review this one.

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
     // obsolete details.
     bool is_obsolete = 9;
   }
+
+  // Windows details get their own format because the information format and model
+  // don't match a normal detail. Specifically windows updates are done as patches,
+  // thus windows vulnerabilities really are a missing package, rather than a package
```
aysylu: nit: capitalize Windows for additional clarity for the reader

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
...
+
+  message WindowsDetail {
+    // Required. The CPE URI in
+    // [cpe format](https://cpe.mitre.org/specification/) in which the
+    // vulnerability manifests. Examples include distro or storage location for
+    // vulnerable jar.
+    string cpe_uri = 1;
+
+    // The name of the vulnerability.
+    string name = 2;
```
aysylu: Are `name` and `description` required fields?

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
...
+
+    // The description of the vulnerability.
+    string description = 3;
+
+    // Required. The names of the KBs which have hotfixes to mitigate this vulnerability.
+    // Note that there may be multiple hotfixes (and thus multiple KBs) that mitigate
+    // a given vulnerability.
```
aysylu: Maybe add a detail explicitly stating our current understanding that the presence of any of the KBs implies the vulnerability is fixed?

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
...
+    repeated KnowledgeBase kbs = 4;
+
+    message KnowledgeBase {
+      // The KB name (generally of the form KB1234)
```
aysylu: Is it always 4 digits? Or is the idea that KB is followed by digits?

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
...
+      string name = 1;
+      // A link to the KB in the windows update catalog
```
aysylu: Is there a link we can add to the Windows update catalog? Also, capitalize "W" please.

proto/v1beta1/vulnerability.proto
```diff
@@ -89,6 +89,38 @@ message Vulnerability {
...
+    // The name of the vulnerability.
+    string name = 2;
+
+    // The description of the vulnerability.
+    string description = 3;
+
+    // Required. The names of the KBs which have hotfixes to mitigate this vulnerability.
+    // Note that there may be multiple hotfixes (and thus multiple KBs) that mitigate
+    // a given vulnerability.
+    repeated KnowledgeBase kbs = 4;
```
aysylu: I like the idea suggested in the design review to add "fixing_" prefix to the field name for clarity. WDYT?

proto/v1beta1/vulnerability.proto
clrprod: Applied s/window/Window to the whole cl.

proto/v1beta1/vulnerability.proto
clrprod: Name is definitely required. Description is not required, but is a good idea.

clrprod: Updated comment for name.

proto/v1beta1/vulnerability.proto
clrprod: Added fixing semantics.

proto/v1beta1/vulnerability.proto
clrprod: It's not always 4 digits (and as the number of KBs goes up it will continue to gain digits).

proto/v1beta1/vulnerability.proto
clrprod: Changed name to fixing_kbs.

proto/v1beta1/vulnerability.proto
aysylu: Gotcha. Maybe then modify this to be KB[0-9]+ or something to prevent the reader from making incorrect assumptions? I'm open to ideas here for how to clarify.

clrprod: Updated comment to include regex.

PR #245, opened 2019-01-15
Switch windows_details proto ID.

clrprod:


proto/v1beta1/vulnerability.proto
```diff
@@ -90,11 +90,13 @@ message Vulnerability {
     bool is_obsolete = 9;
   }
 
+  // ID 4 is unusable due to a proto conflict.
+
```
aysylu: Shall we add a comment with the next ID to use?

PR #248, opened 2019-01-24
Added v1 protos, not yet ready for general use

aysylu:
Currently, these are just straight copies of v1beta1 protos, to unblock some of the internal development. Added appropriate warning at the top of each proto about the lack of readiness of these protos.


proto/v1/attestation.proto
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 The Grafeas Authors. All rights reserved.
```
vtsao: Can you update the copyright to 2019 for all these V1 files.

aysylu: good catch, done, thanks!

PR #254, opened 2019-02-20
Fix doc bugs (broken links and typo)

michelleirvine:
Linking to the top of the Docker reference page rather than anchors to minimize chance that changes on the Docker side will break these links.

Fixed typo: "vulnerabilites" to "vulnerabilities".


proto/v1beta1/swagger/grafeas.swagger.json
```diff
@@ -534,7 +534,7 @@
         "SHELL"
```
vtsao: This file is autogenerated, we don't need to edit it manually.

PR #255, opened 2019-02-21
Adding effective_severity

nms2140:
Remove the distro specific severity field from Package Issue and add a new field named effective_severity to the vulnerability for v1 protos.


proto/v1/vulnerability.proto
```diff
@@ -120,6 +120,11 @@ message Details {
 
   // Output only. URLs related to this vulnerability.
   repeated grafeas.v1.RelatedUrl related_urls = 7;
+
+  // The distro assigned severity for this vulnerability when that is
```
aysylu: nit: s/that/it

proto/v1/vulnerability.proto
```diff
@@ -120,6 +120,11 @@ message Details {
...
+  // available and note provider assigned severity when distro has not yet
```
aysylu: nit: comma after 'available'

proto/v1/vulnerability.proto
```diff
@@ -120,6 +120,11 @@ message Details {
...
+  // assigned a severity for this vulnerability.
+  Severity effective_severity = 8;
```
vtsao: @aysylu any thoughts on the name of this field?

aysylu: I'm fine with `effective` in the name, or `distro_severity`. Is it important for us to distinguish whether distro has already assigned severity for a vulnerability, or whether we're using a note provider's value?

nms2140: I would prefer not to call it distro_severity, since it will be populated with the severity from cvss scores when distro severity is unavailable. I believe it is ok for us to not distinguish which source of the severity we are using to populate the field.

proto/v1/vulnerability.proto

PR #260, opened 2019-03-08
Reserve tag 4 in vulnerability.proto

noahklein:


proto/v1/vulnerability.proto
```diff
@@ -94,6 +94,7 @@ message Vulnerability {
   }
 
   // ID 4 is unusable due to a proto conflict.
+  reserved 4;
```
tylerdatgoogle: Please move this to the top of the message for easier visibility. We eventually would want all reserved proto fields in the same place.

proto/v1beta1/vulnerability.proto
```diff
@@ -91,6 +91,7 @@ message Vulnerability {
   }
 
   // ID 4 is unusable due to a proto conflict.
+  reserved 4;
```
tylerdatgoogle: same comment

proto/v1/vulnerability.proto
```diff
@@ -45,6 +45,8 @@ enum Severity {
 
 // Vulnerability provides metadata about a security vulnerability in a Note.
 message Vulnerability {
+  reserved 4;
```
aysylu: Please comment this out, but keep the wording.

tylerdatgoogle: Are you saying to comment out the line "reserved 4"? The motive for this PR is to add this tag using the reserved keyword: https://developers.google.com/protocol-buffers/docs/proto#reserved.

aysylu: @tylerdatgoogle and I chatted about this and reached an agreement that we should revert this PR. Sorry for the troubles, @noahklein. My concern here is balancing good OSS stewardship and ensuring compatibility with internal Grafeas implementations. Making this an un-commented `reserved` field would conflict with internal usage and implications for other Grafeas users. Commenting out the `// reserved 4` has the same effect as the state before, so it's a no-op. I filed an issue for myself to look into this more comprehensively. Thank you both!

PR #261, opened 2019-03-08
Removed Discovery's last_analysis_time field as it's not needed.

aysylu:
Short explanation: it doesn't make sense for the Discovery to be analyzed at a subsequent time. So we can just remove this field.

### DIFF
--- a/proto/v1/discovery.proto
+++ b/proto/v1/discovery.proto
@@ -45,8 +45,6 @@ message Details {
 
 // Provides information about the analysis status of a discovered resource.
 message Discovered {
-  reserved 2;
-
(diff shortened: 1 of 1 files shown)


proto/v1/discovery.proto
```diff
@@ -45,6 +45,8 @@ message Details {
 
 // Provides information about the analysis status of a discovered resource.
 message Discovered {
+  reserved 2;
```
vtsao: You can just renumber the tag values since this is a brand new proto that nobody is using.

aysylu: Good point, done.

PR #266, opened 2019-03-27
Fixed govet directive in Makefile

aysylu:


Makefile
```diff
@@ -22,10 +22,10 @@ fmt:
 	@gofmt -l -w $(SRC)
 
 test:
-	@go test -v ./... 
+	@go test -v ./...
 
 vet:
-	@go tool vet ${SRC}
+	@go vet ./...
```
vtsao: Why this change? I think we still want to only vet the ${SRC} since it doesn't include the vendor files right?

aysylu: The `tool` part no longer allows the command to run. The path fix is needed because of [this error](https://stackoverflow.com/questions/53802745/after-go-vet-occurs-named-files-must-all-be-in-one-directory).

PR #268, opened 2019-03-29
Updated vendor directory

aysylu:
Ran `govendor update +v` to update.


aysylu:
> Can you provide a description command you ran for this update?
Done!

PR #269, opened 2019-03-31
Fixed Makefile

aysylu:
1) Fixed the proto generation rules for golang.
2) Fixed the dependencies of Makefile targets.


Makefile
```diff
@@ -21,27 +30,29 @@ build: vet fmt go_protos swagger_docs
...
 	curl https://github.com/google/protobuf/releases/download/v3.3.0/protoc-3.3.0-linux-x86_64.zip -o protoc/protoc.zip -L
 	unzip protoc/protoc -d protoc
 
-CLEAN += protoc
+CLEAN += protoc proto/*/*_go_proto
+
+GO_PROTO_DIRS := $(patsubst %.proto,%_go_proto/.done,$(wildcard proto/*/*.proto))
 
-go_protos: grafeas_go_v1alpha1 proto/v1beta1/*_go_proto proto/v1/*_go_proto
+go_protos: v1alpha1/proto/grafeas.pb.go $(GO_PROTO_DIRS)
```
vtsao: What happened to beta and V1?

vtsao: Oh, is it because alpha has a different path format? Can you leave a comment here explaining that?

vendor/github.com/golang/protobuf/descriptor/descriptor_test.go
```diff
@@ -1,32 +0,0 @@
-package descriptor_test
```
vtsao: How come a lot of these vendor files were removed?

Makefile
aysylu: Added the comment, thanks!

vendor/github.com/golang/protobuf/descriptor/descriptor_test.go
aysylu: Hm, I'll wait for the other PR with vendor changes to go in first, then merge and see if it shows up again.

aysylu: The vendor changes went in with the PR #268.

PR #271, opened 2019-03-31
Add minimalistic helm chart to grafeas

liron-l:
A minimalistic grafeas helm chart to -

The user must provide (see values.yaml)
1. Certificate for mutial authentication
2. Name to a persistent volume claim (to persistent the DB data)

If you are OK with the proposed approach, i would add a README on how to
run it locally.


aysylu:
Hi @liron-l, thanks for the PR. Could you please add more details for this on the intended use? README would definitely be helpful on how to bring this up. For `0.1.0` we're considering using Helm charts for Grafes, to bring it up with Kritis, so I think this PR is on the right track, but I'd like to understand better the approach you're using here.

liron-l:
Thanks @aysylu, i've added a README with instructions on how to run the chart locally.

helm/README.md
```diff
@@ -0,0 +1,55 @@
+# Grafeas HELM chart
+This folder contains a sample helm chart for running grafeas using helm.
+The default setup will run a Greafeas instance backed by local [boltdb](https://github.com/boltdb/bolt) data store  with mutual tls authentication.
```
aysylu: It'd be great if the default instance used the in-memory storage instead. Happy to include the `boltdb` version in a separate repository, `grafeas-boltdb`, similar to [grafeas-pgsql](https://github.com/grafeas/grafeas-pgsql).

aysylu:
Thank you, @liron-l! Looks great, just 1 piece of feedback about the default installation.

liron-l:
Thanks @aysylu, by default, the helm chart will use the memstore storage type and will not require a persistent volume (see storageType: "memstore" in `values.yaml` and conditions in deployment.yaml).
Also, as the embedded storage solution does not require any additional setup (except from the persistent volume) I kept it as an alternative storage option (see README.md).

aysylu:
Sounds great, thanks @liron-l! Sorry I missed the part about the default memstore config, I see it now. Thanks for your valuable contribution!

PR #278, opened 2019-04-17
proto change for canonicalversion

nms2140:
Adding a canonicalVersion string to the Version proto.

### DIFF
--- a/proto/v1/package.proto
+++ b/proto/v1/package.proto
@@ -127,4 +127,8 @@ message Version {
   // Required. Distinguishes between sentinel MIN/MAX versions and normal
   // versions.
   VersionKind kind = 4;
+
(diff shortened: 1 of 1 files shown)


proto/v1/package.proto
```diff
@@ -127,4 +127,8 @@ message Version {
   // Required. Distinguishes between sentinel MIN/MAX versions and normal
   // versions.
   VersionKind kind = 4;
+
+  // Human readable version string. This string is of the form
+  // <epoch>:<name>-<revision> and is only set when kind = NORMAL.
+  string canonicalVersion = 5;
```
vtsao: kind is NORMAL.

I think we should call this full_name.

nms2140:
Changed the name of the string to full_name

PR #279, opened 2019-04-18
Add a fix_available field to vulnerability occurrence details.

nms2140:


proto/v1/vulnerability.proto
```diff
@@ -165,6 +165,11 @@ message Details {
   // available, and note provider assigned severity when distro has not yet
   // assigned a severity for this vulnerability.
   Severity effective_severity = 8;
+
+  // Output only. Set to true if at least one of the package issues
+  // in this message has a fix available (fixed_location is not set to
+  // MAXIMUM)
```
vtsao: I think we can simply to:

  // Output only. True if at least one of the affected packages
  // has a fix available.

PR #280, opened 2019-04-22
Flatten PackageIssue by removing the VulnerabilityLocation message. 

nms2140:


proto/v1/vulnerability.proto
```diff
@@ -171,25 +171,33 @@ message Details {
...
+  string affected_package = 2;
 
-  // Required. The version of the package being described.
-  grafeas.v1.package.Version version = 3;
+  // Required. The affected version of the vulnerability.
+  grafeas.v1.package.Version affected_version = 3;
+
+  // The CPE URI in [cpe format](https://cpe.mitre.org/specification/)
+  // format. Examples include distro or storage location for vulnerable jar.
+  // This field will only be populated if different from affected_cpe_uri.
```
vtsao: I think we should still populate this field even if it's same as the affected CPE.

proto/v1/vulnerability.proto
```diff
@@ -171,25 +171,33 @@ message Details {
...
+  string fixed_cpe_uri = 4;
+
+  // The package in which fix is available. This field will only be populated
+  // if different from affected_package.
```
vtsao: Same.

nms2140:
I have deleted the VulnerabilityLocation message from this proto, adding that back in since it seems like it is used in the Note as well. We are choosing to leave the VulnerabilityLocation in the Note.

nms2140:
Had to update some failing tests etc.

go/v1/api/occurrence_test.go
```diff
@@ -904,14 +904,18 @@ func vulnzOcc(t *testing.T, pID, noteName, imageName string) *gpb.Occurrence {
...
-							Version: &pkgpb.Version{
+						AffectedCpeUri:  "cpe:/o:debian:debian_linux:8",
+						AffectedPackage: "abc",
+						AffectedVersion: &pkgpb.Version{
 								Name: "0.2.0",
 								Kind: pkgpb.Version_NORMAL,
-							},
-						},
+					        },
+                                                FixedCpeUri:  "cpe:/o:debian:debian_linux:8",
```
vtsao: Looks like this needs to be Go formatted?

go/v1/api/validators/vulnerability/vulnerability.go
```diff
@@ -116,19 +116,32 @@ func ValidateDetails(d *vulnpb.Details) []error {
...
-			errs = append(errs, fmt.Errorf("affected_location.%s", err))
+		for _, err := range pkg.ValidateVersion(ver) {
+			errs = append(errs, fmt.Errorf("version.%s", err))
 		}
 	}
 
-	if fl := p.GetFixedLocation(); fl != nil {
-		for _, err := range validateVulnerabilityLocation(fl) {
-			errs = append(errs, fmt.Errorf("fixed_location.%s", err))
+        if p.GetFixedCpeUri() == "" {
```
vtsao: Fixed CPE URI and package are not required right?

vtsao: They are only required when the fixed version is normal.

go/v1/api/validators/vulnerability/vulnerability.go
```diff
@@ -116,19 +116,32 @@ func ValidateDetails(d *vulnpb.Details) []error {
 func validatePackageIssue(p *vulnpb.PackageIssue) []error {
```
vtsao: FixAvailable also needs to be validated.

nms2140: fixAvailable should always have the default value (false) since it is a bool, so I dont think we need validation there.

go/v1/api/occurrence_test.go
nms2140: Done

go/v1/api/validators/vulnerability/vulnerability_test.go
```diff
@@ -282,35 +286,132 @@ func TestValidatePackageIssue(t *testing.T) {
...
-						Name: "1.1.2",
-						Kind: pkgpb.Version_NORMAL,
-					},
-				},
+				AffectedCpeUri: "cpe:/o:debian:debian_linux:7",
+				AffectedPackage: "debian",
+				AffectedVersion: &pkgpb.Version{
+                                                Name: "1.1.2",
+                                                Kind: pkgpb.Version_NORMAL,
+                                },
```
vtsao: formatting?

nms2140: Done.

PR #286, opened 2019-05-06
Proto changes for client library generation

Katee:
* Remove sub packages under `grafeas.v1` (e.g. `grafeas.v1.vulnerability`, `grafeas.v1.attestation`)
* Use `grafeas_proto_proto` as `go_package` everywhere
* Rename `Details` fields to `<KIND>Occurrence`
* Use `<KIND>Note` for note details and flatten for some kinds
* Put `source.proto` in `provenance.proto` (Java client generation encountered an error "io/grafeas/v1/Source.java: Tried to write the same file twice" without this change)
* Remove unused timestamp import from `discovery.proto`
* Add make target for the v1 protos


proto/v1/grafeas.proto
```diff
@@ -208,20 +208,20 @@ message Occurrence {
   // resource.
   oneof details {
     // Describes a security vulnerability.
-    grafeas.v1.vulnerability.Details vulnerability = 8;
+    grafeas.v1.VulnerabilityOccurrence vulnerability = 8;
```
vtsao: Do you think we should we rename all these fields to be [KIND]Occurrence as well?

aysylu: I'd vote for now, for brevity. However, here are suggestions for renaming, to be consistent with the overall naming strategy:
* `derived_image` -> `image`
* `installation` -> `package`
* `discovered` -> `discovery`

proto/v1/grafeas.proto
```diff
@@ -279,19 +279,19 @@ message Note {
...
   oneof type {
     // A note describing a package vulnerability.
-    grafeas.v1.vulnerability.Vulnerability vulnerability = 10;
+    grafeas.v1.VulnerabilityNote vulnerability = 10;
     // A note describing build provenance for a verifiable build.
-    grafeas.v1.build.Build build = 11;
+    grafeas.v1.BuildNote build = 11;
     // A note describing a base image.
-    grafeas.v1.image.Basis base_image = 12;
+    grafeas.v1.ImageNote base_image = 12;
```
aysylu: Ditto here for naming strategy of the fields.

proto/v1/grafeas.proto
Katee: @vtsao do you mean changing all lines to (for example) `grafeas.v1.VulnerabilityOccurrence vulnerability_occurrence = 8;`?

aysylu: Oops, typo, should have been `I'd vote for no`.

Katee:
I force pushed a change to remove the proto re-generation from this PR.

proto/v1/grafeas.proto
vtsao: Yeah this LGTM.

aysylu:
LGTM, thanks for implementing the requested changes.

PR #288, opened 2019-05-09
Changing layer directive to be a string, since the Docker directives …

vtsao:
…can change.


vtsao:
I hate Git.

aysylu:
> I hate Git.

Did you pull from `upstream` and `origin` and then push?

PR #290, opened 2019-05-14
remove GetVulnerabilityOccurrencesSummary

vtsao:
This method is too specific to vuln occurrences.


Katee:
Why does this also include a bunch of files removed from the `v1beta1` folder?

PR #292, opened 2019-05-14
changing Resource composite type to just a resource_uri string

vtsao:
The Resource composite type is confusing and probably not needed for now. Reverting it to a resource_uri string.

### DIFF
--- a/proto/v1/grafeas.proto
+++ b/proto/v1/grafeas.proto
@@ -171,8 +171,10 @@ message Occurrence {
   // `projects/[PROJECT_ID]/occurrences/[OCCURRENCE_ID]`.
   string name = 1;
 
-  // Required. Immutable. The resource for which the occurrence applies.
-  Resource resource = 2;
+  // Required. Immutable. A URI that represents the resource for which the
+  // occurrence applies. For example,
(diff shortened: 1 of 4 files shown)


proto/v1/grafeas.proto
```diff
@@ -171,8 +171,10 @@ message Occurrence {
   // `projects/[PROJECT_ID]/occurrences/[OCCURRENCE_ID]`.
   string name = 1;
 
-  // Required. Immutable. The resource for which the occurrence applies.
-  Resource resource = 2;
```
xingx-xx: Why just use a simple string to replace resource?
I think **Resource** with a **content_hash** field is very useful in cases such as a SnapShot jar or some dependencies accidentally overwritten.

aysylu: cc @vtsao

vtsao: We hadn't fully fleshed out the Resource message and its use cases, so we decided to move back to a resource string. One day we might switch to a Resource message again.

This way the URL is more free form and providers can construct the URL in whichever way they want to represent the resource. For now if you're using content hashes, we recommend putting it in the URL, just like a docker image does.

PR #295, opened 2019-05-17
Flattening vuln note details fixed location into the details. We did …

vtsao:
…this for vuln occ package issues. This makes it easier to read.

Moved def of PackageIssue message into the vuln occurrence, since it's only used there. This matches the vuln note details def.

Cleaned up comments in the vuln proto file.

### DIFF
--- a/proto/v1/vulnerability.proto
+++ b/proto/v1/vulnerability.proto
@@ -25,7 +25,7 @@ import "proto/v1/common.proto";
 import "proto/v1/cvss.proto";
 import "proto/v1/package.proto";
 
-// Note provider-assigned severity/impact ranking.
+// Note provider assigned severity/impact ranking.
 enum Severity {
   // Unknown.
   SEVERITY_UNSPECIFIED = 0;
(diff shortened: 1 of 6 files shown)


proto/v1/vulnerability.proto
```diff
@@ -41,54 +41,61 @@ enum Severity {
...
 
-    // The min version of the package in which the vulnerability exists.
-    grafeas.v1.Version min_affected_version = 3;
+    // Required. The package this vulnerability affects.
+    string affected_package = 5;
 
-    // The severity (eg: distro assigned severity) for this vulnerability.
-    string severity_name = 4;
+    // Required. The minimum version of the package this vulnerability affects.
+    grafeas.v1.Version affected_version = 6;
```
Katee: I think leaving `min_` in this name might be clearer. I could be convinced otherwise.

Katee: Maybe this is better.

vtsao: Added min_ back and also added it to the vuln occurrence's package issue affected location field as well.

PR #297, opened 2019-05-17
[FEATURE] Move to go modules

duhruh:
Hello!

This pr does a couple of things, the sample app itself was behind and I found it to have a lot of issues running the grpc server and the http rest server on the same port. I was getting a lot of

```
grpc: addrConn.transportMonitor didn't get server preface after waiting. Closing the new transport now
```

and 

```
all SubConns are in TransientFailure, latest connection error: <nil> 
```

along with other various errors. I researched this a lot and found other people experiencing similar problems

https://github.com/grpc/grpc-go/issues/2636
https://github.com/soheilhy/cmux/issues/64

Long story short using the cmux approach doesn't seem to be worth it as it cause more problems than it solves. Instead I moved to using https://github.com/grpc-ecosystem/grpc-gateway which is what this pr introduces.

This also moves us to use go modules so we no longer need* to check in the `vendor/` directory. I noticed once I removed it, generation of protos seemed to be broken. So I created a new `make` target to pull in the well-know google types during the build `make google/googleapis && make google/grpc-gateway`. I had to do this because go modules do not download non go package directories see https://github.com/golang/go/issues/26366

We also add a `make certs`  target for local testing.

The last thing this pr does is clean up the `Dockerfile` and add `docker-compose.yaml` for development.

## Testing 

Building protos

```
go mod vendor # downloads go deps to vendor/  you could also run docker-compose run grafeas go mod vendor
make go_protos # builds protos
```

Running sample server

```
create samples/server/go-server/api/server/config.yaml
docker-compose up grafeas
```

Running sample client

```
docker-compose up client
// or
docker-compose up cert // if you're using tls
```

# Notes
We probably want to make the make target `protoc/bin/protoc` be no os specific. I had to change it to be osx. If you guys like this pr i can make this change.

Thanks again for this great project 😸


aysylu:
Hi @duhruh! Thanks for the PR. I'm reviewing this, but the full review will be delayed as I'm traveling for conferences to talk about Grafeas and Kritis, and am working to release `0.1.0`. In the meantime, could you please open a FR referring this PR to explain why the move to `go modules` is necessary (you already cover this in the PR description, so copy/paste is fine). I'd like to give the community an opportunity to take a look and chime in. Also, once I'm back from the travels, I'd like to schedule a community call with you and others in the community interested in this, to understand this proposal better. In the meantime, please stay tuned as `0.1.0` will be impacting the code base quite a bit. Thanks again and looking forward to your contributions and our discussions!

duhruh:
@aysylu thanks for the response! I created a proposal https://github.com/grafeas/grafeas/issues/301 and linked back to this issue. I also look forward to discussing this with the community 😄

clrprod:
It also turns out that the proto we need from cel got moved to https://github.com/googleapis/googleapis/blob/e266a1ca806aac193ee3ed5d6d2031bea351546e/google/api/expr/v1alpha1/syntax.proto

duhruh:
@clrprod i've updated this pr to only be the go modules update and the necessary Dockerfile update to ensure it still builds. I'll add the fixes for the example app once the modules switch happens.

duhruh:
I had to rebase on master and change a couple of things to get it to work, so any questions are more than welcome :)

clrprod:
Looks good - I think we want to keep as much logic in go generate as we can - it's simpler to understand than the make file. For os specific stuff you can use go build tags to run different generates for different architectures (I think). I think we're blocked on aysylu having review cycles.

Makefile
```diff
@@ -35,12 +35,38 @@ test: generate
...
 clean:
 	go clean ./...
 	rm -rf $(CLEAN)
+
+
+CLEAN += google/googleapis google/grpc-gateway
+
+google/googleapis:
+	mkdir -p google
+	curl -sSL https://github.com/googleapis/googleapis/archive/3b943eb373600e969c247017ea05bb4ca62dfd68.zip -o google/googleapis.zip
```
clrprod: I'd move this and below into a generate file and just use go generate.

Makefile
```diff
@@ -35,12 +35,38 @@ test: generate
...
+CLEAN += protoc
+PROTO_URL := https://github.com/protocolbuffers/protobuf/releases/download/v3.7.1/protoc-3.7.1
+protoc/bin/protoc:
+	mkdir -p protoc
+ifeq ($(shell uname),Darwin)
+	curl $(PROTO_URL)-osx-x86_64.zip -o protoc/protoc.zip -L
+else
+	curl $(PROTO_URL)-linux-x86_64.zip -o protoc/protoc.zip -L
+endif
+	cd protoc && unzip protoc.zip -d .
```
clrprod: I think you can do this via go build tags.

clrprod:
#334 has been merged, so this pull request is superseeded - Thanks for taking the initial stab at this :)

PR #298, opened 2019-05-20
remove the "next_id" comments in protos

vtsao:
We follow semantic versioning in Grafeas and breaking changes require a new API version (which means a new set of protos, so we can reset the numbering any way we want), so we don't have to worry about reserving proto tag values.


clrprod:
Might want to split your commit message into two lines.

Issue #310, opened 2019-06-11
Ensure `go vet` works with the current Grafeas setup and add it back


aysylu:
**Is your feature request related to a problem? Please describe.**
Currently, `vet` command in the `Makefile` fails due to the missing files that are generated by the `build` command:

```go/v1beta1/project/project.go:23:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/project_go_proto
Makefile:37: recipe for target 'vet' failed
make: *** [vet] Error 1
aysylu@aysylu:~/go/src/github.com/grafeas/grafeas$ make
go/v1beta1/api/validators/build/build.go:24:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/build_go_proto
go/v1beta1/api/validators/deployment/deployment.go:23:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/deployment_go_proto
go/v1beta1/api/validators/discovery/discovery.go:24:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/discovery_go_proto
go/v1beta1/api/validators/package/package.go:23:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/package_go_proto
go/v1beta1/api/validators/provenance/provenance.go:23:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/provenance_go_proto
go/v1beta1/project/project.go:23:2: no Go files in /usr/local/google/home/aysylu/go/src/github.com/grafeas/grafeas/proto/v1beta1/project_go_proto
Makefile:37: recipe for target 'vet' failed
make: *** [vet] Error 1
```

There were several other incompatibility issues discovered in the Grafeas codebase with `go vet` that haven't yet been addressed.

**Describe the solution you'd like**
Ensure all issues detected by `go vet` are fixed in the Grafeas codebase, and re-enable `vet` in the Makefile: `build: vet fmt go_protos swagger_docs`.

**Describe alternatives you've considered**
For now, the `vet` command is disabled in the `build` command.

**Additional context**
The issue is also reported in https://github.com/grafeas/grafeas/issues/308

aysylu:
[...]

Issue #308, opened 2019-06-11
Build fails locally


drewboswell:
## Expected Behavior
```bash 
# should work without error
make build
docker build .
```
## Actual Behavior
```
Sending build context to Docker daemon  12.67MB
Step 1/9 : FROM golang:1.12.5
 ---> 1ef078f0da9e
Step 2/9 : COPY . /go/src/github.com/grafeas/grafeas/
 ---> a93781e4f733
Step 3/9 : WORKDIR /go/src/github.com/grafeas/grafeas/samples/server/go-server/api/server/main
 ---> Running in 6fa63b365058
Removing intermediate container 6fa63b365058
 ---> 595481c975ec
Step 4/9 : RUN CGO_ENABLED=0 go build -o grafeas-server .
 ---> Running in 5ccc054b7a96
client.go:7:2: cannot find package "github.com/grafeas/grafeas/proto/v1beta1/grafeas_go_proto" in any of:
	/go/src/github.com/grafeas/grafeas/vendor/github.com/grafeas/grafeas/proto/v1beta1/grafeas_go_proto (vendor tree)
	/usr/local/go/src/github.com/grafeas/grafeas/proto/v1beta1/grafeas_go_proto (from $GOROOT)
	/go/src/github.com/grafeas/grafeas/proto/v1beta1/grafeas_go_proto (from $GOPATH)
../api/api.go:30:2: cannot find package "github.com/grafeas/grafeas/proto/v1beta1/project_go_proto" in any of:
	/go/src/github.com/grafeas/grafeas/vendor/github.com/grafeas/grafeas/proto/v1beta1/project_go_proto (vendor tree)
	/usr/local/go/src/github.com/grafeas/grafeas/proto/v1beta1/project_go_proto (from $GOROOT)
	/go/src/github.com/grafeas/grafeas/proto/v1beta1/project_go_proto (from $GOPATH)
The command '/bin/sh -c CGO_ENABLED=0 go build -o grafeas-server .' returned a non-zero code: 1
Makefile:42: recipe for target 'build-docker-image' failed
make: *** [build-docker-image] Error 1
```

## Steps to Reproduce the Problem
clone repo locally, run make or docker build...

vtsao:
@aysylu is this due to https://github.com/grafeas/grafeas/pull/306?

aysylu:
[...]

PR #311, opened 2019-06-11
Temporarily remove `go vet` from the `build` command

aysylu:
Addresses #308 . #310 is filed to investigate and re-enable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ CLEAN += .install.protoc-gen-go .install.grpc-gateway
 .install.grpc-gateway:
 	go get -u -v github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger && touch $@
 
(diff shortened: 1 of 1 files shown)


vtsao:
Can you elaborate on the issues you're running into with vet? Also, what are the issues we run into that require us to set composites to false? It should always be true. Unless protoc generates proto bindings that violate that style rule? Which I don't think it does?

aysylu:
Yes, `composites=false` was set in #266, idea from [this thread](https://stackoverflow.com/questions/36273920/disable-go-vet-checks-for-composite-literal-uses-unkeyed-fields). The issue with `vet` that inspired this PR is described in #310.

vtsao:
I still don't understand what is not using keyed literals that we need to set composites to false for?

Also, let's confirm he's running into a Docker issue and not make. The log he posted is from Docker build command.

aysylu:
Withdrawing the PR for now, as it's not solving the root cause and I won't be able to get to this by the end of the month. I'll follow up with the reason for the vet directive flag directly with @vtsao

Issue #301, opened 2019-05-22
Proposal: Move to go modules


duhruh:
PR: https://github.com/grafeas/grafeas/pull/297

# Thesis
Adding and removing libraries should be easy and shouldn't require us to bloat our PR's with `vendor/` dependancies. Making PR's smaller and more manageable will allow us to prototype, develop, and review much faster.

# Why
The sample app, as it stands right now has many issues running the GRPC and HTTP rest server on the same port. After attempting to run it I discovered many issues.

```
grpc: addrConn.transportMonitor didn't get server preface after waiting. Closing the new transport now
```

and 

```
all SubConns are in TransientFailure, latest connection error: <nil> 
```

along with other various errors. I researched this a lot and found other people experiencing similar problems

https://github.com/grpc/grpc-go/issues/2636
https://github.com/soheilhy/cmux/issues/64

# Implementation 

Long story short using the cmux approach doesn't seem to be worth it as it cause more problems than it solves. Instead I moved to using https://github.com/grpc-ecosystem/grpc-gateway which is what my pr introduces.

This also moves us to use go modules since I felt it unnecessary and cumbersome to remove certain `vendor/` dependancies in favor of others. 

Now we no longer *need* to check in the `vendor/` directory. I noticed once I removed it, generation of protos seemed to be broken. So I created a new `make` target to pull in the well-know google types during the build `make google/googleapis && make google/grpc-gateway`. I had to do this because go modules do not download non go package directories see https://github.com/golang/go/issues/26366

## Testing 

Building protos

```
go mod vendor # downloads go deps to vendor/  you could also run docker-compose run grafeas go mod vendor
make go_protos # builds protos
or
make build
```

Running sample server

```
# create samples/server/go-server/api/server/config.yaml
docker-compose up grafeas
```

Running sample client

```
docker-compose up client
// or
docker-compose up cert // if you're using tls
```

# Notes
We probably want to make the make target `protoc/bin/protoc` not be os specific.

aysylu:
Thank you, @duhruh, for the thorough and clear summary! This sounds reasonable. How do you feel about meeting to discuss this in a community call in July to move forward with the proposal? The reason for having this in July and not sooner is the focus on getting `0.1.0` release out, which is under active work.

duhruh:
[...]

PR #316, opened 2019-06-17
makefile: Add check for makefile location, and pin to version make based binary dependencies.

clrprod:
The pinning means we shouldn't be broken by upstream.

The makefile check makes it a lot quicker to realize you need to put grafeas into a go path.


Makefile
```diff
@@ -11,17 +11,35 @@
...
-	go get -u -v github.com/golang/protobuf/protoc-gen-go && touch $@
+	rm -rf $$GOPATH/src/github.com/golang/protobuf
+	mkdir -p $$GOPATH/src/github.com/golang
+	git clone --branch v1.3.1 https://github.com/golang/protobuf.git $$GOPATH/src/github.com/golang/protobuf
+	go install -v github.com/golang/protobuf/protoc-gen-go
+	touch $@
 
 .install.grpc-gateway:
-	go get -u -v github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger && touch $@
+	rm -rf $$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
```
aysylu: Should this be under `clean`?

Makefile
```diff
@@ -11,17 +11,35 @@
...
+	if [ "$(realpath ${EXPECTED_MAKE})" != "$(realpath $(lastword $(MAKEFILE_LIST)))" ]; \
+	then  \
+	echo "Makefile is not in GOPATH root"; \
+	false; \
+	fi
+
 CLEAN += .install.protoc-gen-go .install.grpc-gateway
 .install.protoc-gen-go:
-	go get -u -v github.com/golang/protobuf/protoc-gen-go && touch $@
+	rm -rf $$GOPATH/src/github.com/golang/protobuf
```
aysylu: Should we move this into `clean`?

Makefile
```diff
@@ -11,17 +11,35 @@
...
+	mkdir -p $$GOPATH/src/github.com/golang
```
aysylu: Why do we need to create this dir?

aysylu:
Partially addresses #301.

PR #321, opened 2019-06-18
makefile: Use go generate for v1alpha1 protos

clrprod:
Note that this is a WIP which currently only does v1alpha1.

Also note that go:generate supports command shorthand, so for beta and
ga, it'll just be

//go:generate protoc foo.proto

for each file, and then a messy proto command at the top.


vtsao:
Want @aysylu to LGTM as well.

aysylu:
Let's submit this in 2 parts: for `v1alpha1` protos (as is), and then a follow-up for `v1beta1` and `v1`. Then, we'll just want to update the name of this PR to match the intention.

aysylu:
I updated the title of the PR to reflect the suggestion and merge in the meantime. @clrprod when you're ready, happy to review the PR with `go generate` for `v1` and `v1beta1` protos.

PR #318, opened 2019-06-18
Add max affected version back into Note Detail.

nms2140:

PR #322, opened 2019-06-18
Rename MinAffectedVersion in PackageIssue to AffectedVersion

nms2140:


vtsao:
Want LGTM from @aysylu as well.

proto/v1/vulnerability.proto
```diff
@@ -73,25 +73,35 @@ message VulnerabilityNote {
...
-    // The package this vulnerability was fixed in. It is possible for this to
-    // be different from the affected_package.
-    string fixed_package = 8;
-
-    // Required. The version of the package this vulnerability was fixed in.
-    // Setting this to VersionKind.MAXIMUM means no fix is yet available.
-    grafeas.v1.Version fixed_version = 9;
+    // Required. The version number at the start of an interval in which this vulnerability
+    // exists. A vulnerability can affect a package between version numbers
+    // that are disjoint sets of intervals (Example: [1.0.0-1.1.0], [2.4.6-2.4.8]
```
aysylu: nit: let's use lower case 'e' here and ditto for the below

proto/v1/vulnerability.proto
```diff
@@ -73,25 +73,35 @@ message VulnerabilityNote {
...
+    // and [4.5.6-4.6.8]).
```
aysylu: Please add: "each of which will be represented in its own Detail"

aysylu: ditto for the `affected_version_end` comment

aysylu: Let's add a clarification that if a specific affection version is provided by a vulnerability database, `affected_version_start` and `affected_version_end` will be the same.

proto/v1/vulnerability.proto
nms2140: Addressed in Pull request #318

proto/v1/vulnerability.proto
nms2140: addressed in PR #318

nms2140: Addressed in PR #318

PR #324, opened 2019-06-19
Updated server with v1beta1 API

aysylu:
This is the server implementation for the `0.1.0` release.

PR #326, opened 2019-06-19
Updated Grafeas Dockerfile with 0.1.0 server

aysylu:
Also, added instructions on bringing up Dockerized Grafeas. Builds on top of #324, so I'll merge it once it's submitted.


aysylu:
> Can a circle ci test be rewritten that does the docker build and makes sure the server works? That would likely catch a lot of silly errors we may introduce.

Good idea. Added a test that does just `docker build .`. I plan to add a more comprehensive integration test as part of the upgrade of the testing infrastructure work soon.

PR #327, opened 2019-06-20
Fix `go vet` to work with current Grafeas setup

rxbchen:
Fixed issue regarding `go vet` [#310](https://github.com/grafeas/grafeas/issues/310)


Makefile
```diff
@@ -41,7 +41,7 @@ test: go_protos
 	@go test -v ./...
 
 vet: go_protos
-	@go vet -composites=false ./...
```
vtsao: By default this is true, just remove the flag.

go/v1beta1/storage/storage_test.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/grafeas/grafeas/go/name"
-	"github.com/grafeas/grafeas/go/v1beta1/api"
```
vtsao: Is this an unused import? If so, remove it.

rxbchen: I think this is being used in `func doTestStorage()`

Makefile
rxbchen: Will do :)

aysylu:
It looks like all the requested changes by @vtsao have been addressed, but let's wait for him to take another look and explicitly approve. Thanks for addressing this, @rxbchen!

vtsao:
LGTM

PR #329, opened 2019-06-21
make affected_version_{start,end} and fixed_version fields optional

nms2140:
Currently, we set the minimum version to always be Version.MINIMUM, which means that all versions below fixed version are vulnerable. This is not always the case. While we can set the min version from the NVD, there are a few cases of CVEs where the minimum version of CVEs is not available from the NVD data. In such cases, setting Version.MINIMUM as affected_version_start is incorrect, since it implies that all versions below affected_version_end are vulnerable (while the truth is that we do not know which versions are vulnerable). 

The same case applies to affected_version_end as well as fixed_location. There are cases where they are unknown, and in such cases we must not fill those fields. 

Therefore, we are marking affected_version_start, affected_Version_end and fixed_location as non required fields.


vtsao:
FYI @aysylu

PR #333, opened 2019-06-25
Go generate

clrprod:
Convert more files over to go generate.


protoc/generate.go
```diff
@@ -0,0 +1,7 @@
+//go:generate rm -rf include
```
aysylu: What's the purpose of these commented out lines?

clrprod: That's how you configure go generate. It runs each line in order from within the folder.

noahklein: I think a comment at the top explaining what these are generating, would be helpful. e.g. "Generates the protos and proto compilation tools."

aysylu: Anything that explains that these commented-out lines actually have an effect (and optionally, adding what effect, at high-level), across all generate.go files would be helpful. Copy/paste is fine.

clrprod: I added a comment to each of the generate.go files, let me know if you want more explanation.

PR #336, opened 2019-06-26
Moved v1alpha1 proto and swagger spec

aysylu:
to proto/v1alpha1 for consistency with the rest of the versions


aysylu:
The tests are failing in test util that's deleted in #335, so I'll just re-run the tests once the other one is merged.

proto/v1alpha1/generate.go
```diff
@@ -1,4 +1,11 @@
 // This file generates the v1alpha1 proto bindings for go.
 
-//go:generate ../../protoc/bin/protoc -I ./ -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway -I ../../vendor/github.com/googleapis/googleapis --go_out=plugins=grpc:. --grpc-gateway_out=logtostderr=true:. --swagger_out=logtostderr=true:. grafeas.proto
-package v1alpha1
+//go:generate -command protoc ../../protoc/bin/protoc -I ../../ -I ./ -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway -I ../../vendor/github.com/googleapis/googleapis --go_out=plugins=grpc,paths=source_relative:.  --grpc-gateway_out=logtostderr=true,paths=source_relative:.
+//go:generate -command swagger ../../protoc/bin/protoc -I ../../ -I ./ -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I ../../vendor/github.com/grpc-ecosystem/grpc-gateway -I ../../vendor/github.com/googleapis/googleapis --swagger_out=logtostderr=true:.
+//go:generate rm -rf swagger
+//go:generate mkdir swagger
+
+//go:generate protoc grafeas.proto
```
clrprod: I think all the other proto generators put the output .pb.go in a foo_go_proto folder. I have no strong feelings about doing it, but it may be worth being consistent.

aysylu: Yes, I went back and forth on this and decided to keep as close to the way v1alpha1 did generation as possible. I'm not sure who's still relying on the previous structure, and given that there's only 1 proto, it seemed like an overkill to have an extensive dir structure. I'll keep this as is for now, and consider making it even more consistent in the future.

PR #349, opened 2019-07-19
Create source timestamp fields.

clrprod:
This is used to indicate when the source last changed, for latency
considerations and because CreateTime and UpdateTime are grafeas
specific.

### DIFF
--- a/proto/v1/vulnerability.proto
+++ b/proto/v1/vulnerability.proto
@@ -24,6 +24,7 @@ option objc_class_prefix = "GRA";
 import "proto/v1/common.proto";
 import "proto/v1/cvss.proto";
 import "proto/v1/package.proto";
+import "google/protobuf/timestamp.proto";
 
 // Note provider assigned severity/impact ranking.
 enum Severity {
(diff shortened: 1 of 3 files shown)


proto/v1/vulnerability.proto
```diff
@@ -108,6 +109,10 @@ message VulnerabilityNote {
     // Whether this detail is obsolete. Occurrences are expected not to point to
     // obsolete details.
     bool is_obsolete = 11;
+
+    // The time this information was last changed at the source. This is an
+    // upstream timestamp from the underlying information source.
```
aysylu: Let's include examples of the information source at the end. Ditto for the below.

clrprod: Added an example of the Ubuntu Security Tracker

clrprod:
Also added the proto for v1beta1 since this isn't a v1 only feature.

proto/v1/vulnerability.proto
```diff
@@ -108,6 +109,11 @@ message VulnerabilityNote {
     // Whether this detail is obsolete. Occurrences are expected not to point to
     // obsolete details.
     bool is_obsolete = 11;
+
+    // The time this information was last changed at the source. This is an
+    // upstream timestamp from the underlying information source - i.e. Ubuntu
```
aysylu: nit: s/i.e./e.g.

clrprod: Done.

PR #350, opened 2019-07-19
Fixes location of the helm release script in Dockerfile

aysylu:


aysylu:
> Any way to add a test for this?

Good point, added the tests for this and re-ordered to ensure `docker build` runs before unit tests.

PR #353, opened 2019-07-23
Attempt to fix the CloudBuild `docker build` of `grafeas-charts/Dockerfile`

aysylu:
It's building correctly locally, but not on CloudBuild, so it's not possible to reproduce the issue in my local environment. Attempting a fix to get the paths to work correctly.

### DIFF
--- a/grafeas-charts/Dockerfile
+++ b/grafeas-charts/Dockerfile
@@ -14,7 +14,7 @@
 
 FROM gcr.io/gcp-runtimes/ubuntu_16_0_4 as runtime_deps
 
-RUN apt-get install --no-install-recommends --no-install-suggests -y \
+RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
         git
 
 ENV HELM_VERSION v2.8.1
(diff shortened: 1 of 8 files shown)


grafeas-charts/Dockerfile
```diff
@@ -22,9 +22,9 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}
     tar -zxvf helm-${HELM_VERSION}-linux-amd64.tar.gz && \
     mv linux-amd64/helm /usr/local/bin/helm
 
-WORKDIR /go/src/github.com/grafeas/grafeas/grafeas-charts
-
 COPY . .
 
-RUN chmod +x helm_release.sh
+RUN echo $PWD
```
clrprod: These look like debugging commands - are they still needed?

grafeas-charts/Dockerfile
```diff
@@ -22,9 +22,9 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}
...
+RUN ls $PWD
+RUN chmod +x ./helm_release.sh
```
clrprod: This should be persisting from the outside repo (and git should store execute bits correctly). Is it not working?

grafeas-charts/Dockerfile
aysylu: Yes, I'm trying to figure out why it's not building correctly on CloudBuild, hence added these debugging commands. I'll remove them once I resolve the issue.

grafeas-charts/Dockerfile
aysylu: It didn't seem to be working, but at this point I'm unsure if it's because it couldn't find the script itself or it couldn't actually execute it (the error was ambiguous to me, and I didn't store the exact wording). I'll remove whatever is unneeded from the image once I get CloudBuild to build the image correctly.

PR #371, opened 2019-09-03
Adding infrastructure that will allow us to support filtering for list methods

vtsao:


go/v1/filter/filter.go
```diff
@@ -0,0 +1,44 @@
...
+// ListOccsFilterFn is a function that efficiently handles a specific filter pattern. It parses the
+// specified filter to determine if it can handle it, and if it can, it returns only occurrences
+// that match the filter, and the next page token (if available). The function always returns a bool
+// indicating whether it can answer the specified filter.
+type ListOccsFilterFn func(ctx context.Context, projID, filter, pageToken string, pageSize int32) ([]*gpb.Occurrence, string, bool, error)
+
+// ListOccsFilterer holds functions on how to handle various filter patterns for listing
+// occurrences.
+type ListOccsFilterer struct {
+	Log grafeas.Logger
```
clrprod: Do we not have a top level logging package that we can log to directly? Why do we need to pass this as an object?

vtsao: To allow for different logger impls. Internally we have a different one.

Issue #380, opened 2019-09-16
Support DynamoDb as storage backend


john-tipper:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to use DynamoDB as a storage backend.  I don't want to pay for an always-on database and for this database to be highly-available.  I'd like to use a cloud-native solution.

**Describe the solution you'd like**
AWS DynamoDB should be a supported storage instance.

**Describe alternatives you've considered**
There are no cloud-native storage solutions supported by Grafeas at present.

john-tipper:
I am currently working on an implementation for this issue.

aysylu:
Great! Looking forward to your contribution @john-tipper. Please note that going forward, we'd like to have storage bindings in separate projects, see WIP in [grafeas-pgsql](https://github.com/grafeas/grafeas-pgsql). See [Grafeas architecture slide](https://www.slideshare.net/aysylu/q-con-sp-software-supply-chain-management-with-grafeas-and-kritis/102) for more details.

john-tipper:
Yep, sorry, should have mentioned I've seen https://github.com/grafeas/grafeas/issues/341 and will do similar.  Will be asking for creation of grafeas/grafeas-dynamodb repo when I've got something that works.

aysylu:
@john-tipper no problem and glad to hear we're on the same page! Sounds good. I was also thinking that we can link to the dedicated storage binding projects from the Grafeas main page, but the `grafeas-dynamodb` itself doesn't have to live in the `grafeas` organization, thereby simplifying the governance and requirements specific to contributing to Grafeas. WDYT of this approach?

john-tipper:
Hey @aysylu, I have made a PR (https://github.com/grafeas/grafeas/pull/382) that separates out implementation-specific configuration from the project configuration.  I agree with you: it'll allow for this Grafeas repo to refer in the README to other storage implementations (if you think they're good enough), but there should be no need for further changes inside this repo.  It should make #341 much easier to implement.

I've put some detailed commentary in the PR, let me know what you think - it's my first attempt at Go.

john-tipper:
Initial commit of #380 is at https://github.com/john-tipper/grafeas-dynamodb.  It will require this PR to be merged to work correctly without manually rejigging the Go modules defined in that project.  

It provides a good example for how to use the changes made in #383 to support a new storage system and provide a server implementation that makes use of that implementation, which may well be of interest for anyone working on #341.

john-tipper:
[...]

Issue #341, opened 2019-07-06
Move PostgreSQL storage implementation to `grafeas-pgsql`


aysylu:
**Is your feature request related to a problem? Please describe.**
For consistency with the [Grafeas architecture](https://www.slideshare.net/aysylu/q-con-sp-software-supply-chain-management-with-grafeas-and-kritis/102?src=clipshare), we'd like to move the PostgreSQL storage implementation to its own project, `grafeas-pgsql`.

**Describe the solution you'd like**
1. Move [pgsqlstore.go](https://github.com/grafeas/grafeas/blob/master/go/v1beta1/storage/pgsqlstore.go) to `grafeas-pgsql`.
1. Re-build the Grafeas server with the PostgreSQL storage.

**Describe alternatives you've considered**
For now, it's part of the Grafeas project, but going forward we'd like to keep the project scope limited to the API.

judavi:
I can help with this issue

aysylu:
That'd be fantastic, thank you, @judavi!

judavi:
I can resume the work on this issue based on the experience that I got working on [Grafeas-oracle](https://github.com/judavi/grafeas-oracle)

aysylu:
That'd be great, thanks, @judavi!

meyerkeith:
Wondering if this has moved much more? I'm looking for a way to jump in and help with the codebase a bit more.

PR #382, opened 2019-09-19
#380 support multiple storage implementations

john-tipper:
At present, the configuration of the server includes pointers to all implementation-specific configuration classes (e.g. Postgres, embedded).  As support for new storage types are added, this current model would require changes to this repo to add new storage classes, which is not wanted as we want to be able to develop independently.  This PR is a non-breaking change for current configurations to enable support for adding multiple storage implementations without further changes to this repo.  This will make the implementation of #341 and #380 much easier.

This PR removes implementation-specific configuration from the main configuration class and replaces it with a generic configuration class.  Configuration is parsed from a file using Viper, in order to make use of its capabilities to unmarshal the parsed configuration into an implementation-specific configuration class.  The user can add new storage types by simply registering a storage provider for that new storage implementation.  This approach also allows for new storage implementations to support multiple separate storage implementations in other projects (e.g. a new project could support Spanner and Dynamodb, but not Postgres).

The GRPC initialisation that was previously in the main file has been extracted into utility methods in order to keep the main class as small as possible.  Now the server startup looks like:

```
func main() {
	err := storage.RegisterDefaultStorageTypeProviders()
	if err != nil {
		log.Panicf("Error when registering storage type providers, %s", err)
	}
	StartGrafeas()
}
```

To define a new storage implementation, you can now do this:
```
func main() {
	err := storage.RegisterDefaultStorageTypeProviders()
	if err != nil {
		log.Panicf("Error when registering storage type providers, %s", err)
	}

        // register a new storage type using the key 'mynewstorage'
	err = storage.RegisterStorageTypeProvider("mynewstorage", mnsStorageTypeProvider)

	if err != nil {
		log.Panicf("Error when registering my new storage, %s", err)
	}

	StartGrafeas()
}

func mnsStorageTypeProvider(storageType string, storageConfig *interface{}) (*Storage, error) {
	if storageType != "mynewstorage" {
		return nil, errors.New(fmt.Sprintf("Unknown storage type %s, must be 'mynewstorage'", storageType))
	}

        // some struct config.MyNewStorageConfig is defined elsewhere however you want
	var storeConfig config.MyNewStorageConfig

        // provided your yaml config matches config.MyNewStorageConfig, you should not have an error here
	err := config.ConvertGenericConfigToSpecificType(storageConfig, &storeConfig)
	if err != nil {
		return nil, errors.New(fmt.Sprintf("Unable to create config.MyNewStorageConfig, %s", err))
	}

	s := NewMyNewStorageConfig(&storeConfig)
	storage := &Storage{
		ps: s,
		gs: s,
	}

	return storage, nil
}
```

Your new server implementation now supports all currently supported storage (`memstore`, `embedded` and `postgres`), plus `mynewstorage`.  When #341 is implemented, the new project will support `memstore`, `embedded` and `mynewstorage` urless the new project is amended to add support for Postgres back into it.

Additionally, this PR adds testing for the configuration parsing and confirms that the example configuration given will parse without error.

If Spanner is used internally within Google for Grafeas storage, then it should be trivial to add support for it using this pattern, so this PR should not break any existing configurations.


john-tipper:
@googlebot I signed it!

john-tipper:
Closing as I couldn't get Googlebot to recognise my email address on my commits through rebasing as there was a merge commit in it to update my fork (my Git-fu is lacking here), so have re-opened as PR #383.

PR #391, opened 2019-10-15
Add in-toto attestations to Grafeas

adityasaky:
This PR introduces in-toto attestations for Grafeas. It sets up in-toto links as a type of Grafeas note. We tested it works by running an instance of the server and registering test in-toto notes.


adityasaky:
@googlebot I signed it!

adityasaky:
Hey @aysylu! Got started on defining the proto for in-toto attestations in Grafeas. Does this look like what we discussed?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,41 @@
...
+}
+
+message Signed {
+  string _type = 1;
+
+  string name = 2;
+
+  repeated string command = 3;
+
+  map<string, ArtifactHashObject> materials = 4;
```
aysylu: Could you add more information for each of these fields, `materials`, `products` etc with clarifications on what they represent?

adityasaky: I add comments for all the fields and included a link to the specification which goes into greater detail.

adityasaky:
HI @aysylu, we spent some time yesterday on this branch to get Grafeas in a state to accept in-toto link files as a type of notes. We've tested this with some sample metadata and it looks to be working okay! I also rebased off v0.1.4.

zmarouf:
@adityasaky I'm very interested in in-toto + Grafeas.
Anything I should know about running this?
I imagine this work supersedes https://github.com/in-toto/totoify-grafeas or are they meant to be used in conjunction?

adityasaky:
Hi @zmarouf, yes this supersedes that repository.

This PR is built off v0.1.4 of Grafeas. It introduces in-toto link metadata as a type of Grafeas note. You can clone this branch and get Grafeas server up and running the usual way as described here: https://github.com/grafeas/grafeas/blob/master/docs/running_grafeas.md#using-dockerfile

Here's an example in-toto note we created to test this. The `intoto` field just contains an in-toto link file.

```
{
    "name" : "projects/provider_example/notes/test",
    "shortDescription": "A brief description of the note",
    "longDescription": "A longer description of the note",
    "kind": "INTOTO",
    "intoto": {

        "signatures": [
            {
                "keyid": "776a00e29f3559e0141b3b096f696abc6cfb0c657ab40f441132b345b08453f5",
                "sig": "1cc245803ed63335b95b91a527f740fe176109ffb372c6f0e0075fa069fb9e8d85df4c4be1a33dcce8f5e7d660b14a75e409ee30c43af303e751d38b0697e4cd3cdc586219f638513fd8471bcdeaf762e51a7cd033780da98f364cc39f188faa7d0165f44ba2f051d04d28dab5f7a295876f9fafc31d3bbf91175969f7965b4927d040957cd3f7a6bf002f9fa0eb979b5ba9bf2328fad11be365ddfae158cb35e2f5545da9765499328a9df6faab00993ef7ec0c3f7a7518d28ec41c72c738b2d0eb2da5fb9d1d5f1e01632163117138968acbc2959cf8534b89bb27b9a82f1931a1c6d3533cefcff0ad60aff909eb6de375d71e88d42e3f7504393e2539946a03bb22a9502d12ea2444fc55bb102569f29d02028109a21c85cc1500cef313eb8f13f0339297c3b9eeab9f8350ab15b44029630ee6279dd60c4225dc08833e5dc04a32a27b3c6de6fd8e3b77a77420c0c9a9057cecabfef08bdb34fa5460a1d1c8faedc18d0df20d1cb824a045690d183d25575efdd4019e2aac087948e02d13"
            }
        ],
        "signed": {
            "_type": "link",
            "byproducts": {
                "return-value": 1,
                "stderr": "",
                "stdout": ""
            },
            "command": [
                "git",
                "clone",
                "https://github.com/in-toto/demo-project.git"
            ],
            "environment": {},
            "materials": {},
            "name": "clone",
            "products": {
                "demo-project/foo.py": {
                    "sha256": "ebebf8778035e0e842a4f1aeb92a601be8ea8e621195f3b972316c60c9e12235"
                }
            }
        }
    }
}
```

The Grafeas server that you're running should accept the note for a project if you follow the steps here: https://github.com/grafeas/grafeas/blob/master/docs/create_notes_occurrences.md#creating-a-project and https://github.com/grafeas/grafeas/blob/master/docs/create_notes_occurrences.md#using-curl.

Let me know if you run into any issues! :)

adityasaky:
Hi @aysylu, I've removed the draft status from this PR, I think it's ready for further review and discussion.

Thanks!

skelterjohn:
Thanks for the work! The holiday break is clearing up now and we'll get to this PR soon.

judavi:
@skelterjohn @adityasaky I'll happy to help if something else is needed for complete this integration. I'll love to see in-toto working with Grafeas!

judavi:
@adityasaky I'm wondering a couple of things:
1. Why not store the .links (https://github.com/in-toto/docs/blob/v0.9/in-toto-spec.md#44-file-formats-namekeyid-prefixlink) as occurrences?
The note names should be unique and the occurrences are instances of a note and in this case, the note could be the specific step from the layout.
Also, I'm imaging the scenario where you execute your pipeline.  On every build, you will execute the same step (note) on each execution and then you will have multiple instances of that step (occurrences/.links). Also if you need to query you will see all the occurrences for the specific step under the note.
2. Is the layout from in-toto also considered to be stored in Grafeas?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
+syntax = "proto3";
```
aysylu: Could you please add a header here like in other protos?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+
+// This corrresponds to the core in-toto link attestation.
+message Link {
+  // This field denotes the type of metadata being handled - in this case, the
+  // value is always "link".
+  string _type = 1 [json_name="_type"];
+
+  // This field contains the name of the step the link attestation was generated
+  // for.
+  string name = 2 [json_name="name"];
```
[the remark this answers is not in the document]
aysylu: or `step_name`?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+// This file defines the proto that matches the schema of an in-toto link
+// file. in-toto is a framework to protect a software supply chain. You can read
+// more about in-toto at https://in-toto.io/ and you can find the current
+// version of the in-toto specification at
+// https://github.com/in-toto/docs/blob/master/in-toto-spec.md
+
+
+// A signature object consists of the KeyID used and the signature itself.
+message Signature {
+  string keyid = 1 [json_name="keyid"];
```
aysylu: Remind me: is `json_name` to be able to convert into in-toto format more easily?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+message Link {
+  // This field denotes the type of metadata being handled - in this case, the
+  // value is always "link".
+  string _type = 1 [json_name="_type"];
+
+  // This field contains the name of the step the link attestation was generated
+  // for.
+  string name = 2 [json_name="name"];
+
+  // This field contains the full command executed for the step. This can also
```
aysylu: "the full command executed for the step" -- is it expected that each part of the command will be a separate string in the repeated field? If so, it may be worth clarifying this and maybe adding an example

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+  // be empty if attestations are generated for operations that aren't directly
+  // mapped to a specific command.
+  repeated string command = 3 [json_name="command"];
+
+  // Materials are the supply chain artifacts that go into the step and are used
+  // for the operation performed.
```
aysylu: So, what's the expected string to struct mapping here? It'd be helpful to clarify in the comment.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // Products are the supply chain artifacts generated as a result of the step.
+  map<string, google.protobuf.Struct> products = 5 [json_name="products"];
+
+  // By Products are made up of the data that is generated as part of a software
+  // supply chain step, but are not the actual result of the step.
+  ByProductObject byproducts = 6 [json_name="byproducts"];
+
+  // This is a field that can be used to capture information about the
+  // environment.
```
aysylu: it'd be good to include an example or two for what kind of environment you mean

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,72 @@
...
+  map<string, string> environment = 7 [json_name="environment"];
+}
+
+// This corresponds to a full signed attestation - it is made up of one or more
+// signatures and the core link attestation itself.
+message InToto {
```
aysylu: In `v1` Grafeas API, we made an improvement to make it more obvious which are notes and occurrences, following Grafeas concepts. Is `InToto` in this case a Note?

proto/v1beta1/grafeas.proto
```diff
@@ -293,9 +294,11 @@ message Note {
     grafeas.v1beta1.discovery.Discovery discovery = 15;
     // A note describing an attestation role.
     grafeas.v1beta1.attestation.Authority attestation_authority = 16;
+    // A note describing an in-toto link.
+    grafeas.v1beta1.intoto.InToto intoto = 17;
```
aysylu: We also want to define an intoto occurrence, as in https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/grafeas.proto#L229

proto/v1beta1/intoto.proto
adityasaky: Yes, each part of the command is a separate string. I'll include the example.

```"command": ["git","clone","https://github.com/in-toto/demo-project.git"]```

File: proto/v1beta1/intoto.proto (hunk 8 above)
Comment (adityasaky) [reply proto/v1beta1/intoto.proto]: That's a good point, and ties into what @judavi mentioned above. I updated the definition to accept most of the actual `.link` files as occurrences, which map to a note that contains the step name. I also included the signing keys used by the functionary to sign the link metadata in the note - this is defined in the in-toto layout.

File: proto/v1beta1/intoto.proto (hunk 4 above)
Comment (adityasaky) [reply proto/v1beta1/intoto.proto]: `json_name` in this instance is able to easily map to the in-toto naming of the fields.

**Comment by adityasaky** [comment]
@judavi 

1. I just updated the PR to handle in-toto occurrences. The note now includes the name of the step and a list of public keys that can be used to verify the signatures on the links. It can be further updated to include the artifact rules from the layout for the corresponding step. The note now looks like:
```
{
  "name": "projects/provider_example/notes/test",
  "shortDescription": "A brief description of the note",
  "longDescription": "A longer description of the note",
  "kind": "INTOTO",
  "intoto": {
    "step_name": "clone",
    "signing_keys": [
      {
        "keyid": "776a00e29f3559e0141b3b096f696abc6cfb0c657ab40f441132b345b08453f5",
        "keyid_hash_algorithms": [
          "sha256",
          "sha512"
        ],
        "keytype": "rsa",
        "keyval": {
          "private": "",
          "public": "-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEA0Zfzonp3/FScaIP+KKuz\nB+OZNFpjbVGWjm3leqnFqHYLqrLcCw5KhlXpycJqoSvZBpO+PFCksUx8U/ryklHG\nVoDiB84pRkvZtBoVaA4b4IHDIhz1K5NqkJgieya4fwReTxmCW0a9gH7AnDicHBCX\nlzMxqEdt6OKMV5g4yjKaxf8lW72O1gSI46GSIToo+Z7UUgs3ofaM5UFIcczgCpUa\n5kEKocB6cSZ9U8PKRLSs0xO0ROjrcOTsfxMs8eV4bsRCWY5mAq1WM9EHDSV9WO8g\nqrRmanC4enNqa8jU4O3zhgJVegP9A01r9AwNt6AqgPSikwhXN/P4v1FMYV+R6N3b\nS1lsVWRAnwBq5RFz5zVvcY88JEkHbrcBqP/A4909NXae1VMXmnoJb4EzGAkyUySB\na+fHXAVJgzwyv3I48d/OIjH8NWcVmM/DQL7FtcJk3tp0YUjY5wNpcbQTnLzURtlU\nsd+MtGuvdlDxUUvtUYCIVKRdS8UzYnTPjI2xzeoSHZ2ZAgMBAAE=\n-----END PUBLIC KEY-----"
        },
        "scheme": "rsassa-pss-sha256"
      }
    ]
  }
}
```

And the occurrence looks like:

```
{
  "name": "projects/occurrence_example/occurrences/test",
  "resource": {
    "uri": "some-uri"
  },
  "noteName": "projects/provider_example/notes/testNote",
  "kind": "INTOTO",
  "intoto": {
    "signatures": [
      {
        "keyid": "776a00e29f3559e0141b3b096f696abc6cfb0c657ab40f441132b345b08453f5",
        "sig": "1cc245803ed63335b95b91a527f740fe176109ffb372c6f0e0075fa069fb9e8d85df4c4be1a33dcce8f5e7d660b14a75e409ee30c43af303e751d38b0697e4cd3cdc586219f638513fd8471bcdeaf762e51a7cd033780da98f364cc39f188faa7d0165f44ba2f051d04d28dab5f7a295876f9fafc31d3bbf91175969f7965b4927d040957cd3f7a6bf002f9fa0eb979b5ba9bf2328fad11be365ddfae158cb35e2f5545da9765499328a9df6faab00993ef7ec0c3f7a7518d28ec41c72c738b2d0eb2da5fb9d1d5f1e01632163117138968acbc2959cf8534b89bb27b9a82f1931a1c6d3533cefcff0ad60aff909eb6de375d71e88d42e3f7504393e2539946a03bb22a9502d12ea2444fc55bb102569f29d02028109a21c85cc1500cef313eb8f13f0339297c3b9eeab9f8350ab15b44029630ee6279dd60c4225dc08833e5dc04a32a27b3c6de6fd8e3b77a77420c0c9a9057cecabfef08bdb34fa5460a1d1c8faedc18d0df20d1cb824a045690d183d25575efdd4019e2aac087948e02d13"
      }
    ],
    "signed": {
      "_type": "link",
      "byproducts": {
        "return-value": 1,
        "stderr": "",
        "stdout": ""
      },
      "command": [
        "git",
        "clone",
        "https://github.com/in-toto/demo-project.git"
      ],
      "environment": {},
      "materials": {},
      "name": "clone",
      "products": {
        "demo-project/foo.py": {
          "sha256": "ebebf8778035e0e842a4f1aeb92a601be8ea8e621195f3b972316c60c9e12235"
        }
      }
    }
  }
}
```

2. The layout in itself isn't being stored in Grafeas.

File: proto/v1beta1/intoto.proto (hunk 4 above)
Comment (aysylu) [reply proto/v1beta1/intoto.proto]: Gotcha, is the intention for these fields to be identically named though? It seems that if the field in proto has the same name as in json, `json_name` is unnecessary.

Also, it's more idiomatic in protos to separate the words in snake case, eg `key_id`.

**Comment by aysylu** [comment]
Thanks for the clarifications in the above, @adityasaky! For resource URIs in Grafeas, this could be a final artifact (e.g. container built using the specified layout), or any of the intermediate artifacts.

File: docs/grafeas_concepts.md [hunk 10]
```diff
@@ -86,3 +86,4 @@ the type of information each kind of note and occurrence contains.
 |IMAGE        |Information about the base image of a container|Information about layers included on top of the base image in a particular container|
 |PACKAGE      |Package descriptions                           |Filesystem locations detailing where the package is installed in a specific resource|
 |VULNERABILITY|CVE or vulnerability description and details including severity, versions|Affected packages/versions in a specific resource|
+|INTOTO|An in-toto step|Details of a particular in-toto link attestation of the corresponding step|
```
aysylu: Great! Let's also add a link to in-toto documentation here?

proto/v1beta1/common.proto
```diff
@@ -39,6 +39,8 @@ enum NoteKind {
   DISCOVERY = 6;
   // This represents a logical "role" that can attest to artifacts.
   ATTESTATION = 7;
+  // This represents an in-toto link file.
```
aysylu: This comment is a bit too specific in in-toto terminology. I recommend to expand on what "in-toto link file" means in terms of software supply chain metadata.

proto/v1beta1/grafeas.proto
```diff
@@ -219,9 +220,11 @@ message Occurrence {
     grafeas.v1beta1.discovery.Details discovered = 13;
     // Describes an attestation of an artifact.
     grafeas.v1beta1.attestation.Details attestation = 14;
+    // Describes a specific in-toto InTotolink attestation.
```
aysylu: nit: `InTotolink` formatting, inconsistent with the below

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+// https://github.com/in-toto/docs/blob/master/in-toto-spec.md
+
+
+// A signature object consists of the KeyID used and the signature itself.
+message Signature {
+  string keyid = 1 [json_name="keyid"];
+  string sig = 2 [json_name="sig"];
+}
+
+// An object defined to capture the by products of a step.
```
aysylu: It'd be more helpful to the reader to unpack what "by products of a step" means, and possibly point to in-toto's definition/description of this.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+message ByProductObject {
+  int32 return_value = 1 [json_name="return-value"];
+  string stderr = 2 [json_name="stderr"];
+  string stdout = 3 [json_name="stdout"];
+}
+
+// This corrresponds to the core in-toto link attestation.
```
aysylu: nit: co*rrr*esponds triple "r" :)

aysylu: Does the term "core" mean something specific in in-toto terminology? Also, having a comment explain that this is a Note in Grafeas terminology would be helpful here.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  // string in the list. An example of a command in the in-toto metadata field
+  // is:
+  // "command": ["git", "clone", "https://github.com/in-toto/demo-project.git"]
+  repeated string command = 1 [json_name="command"];
+
+  // Materials are the supply chain artifacts that go into the step and are used
+  // for the operation performed. The key of the map is the path of the artifact
+  // and the structure contains the recorded hash information. An example is:
+  // "materials": {
+  //   "foo/bar": {
```
aysylu: Could we change this to one of the formats described [here](https://github.com/grafeas/grafeas/blob/master/docs/grafeas_concepts.md#resource-urls)?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  // be empty if attestations are generated for operations that aren't directly
+  // mapped to a specific command. Each term in the command is an independent
+  // string in the list. An example of a command in the in-toto metadata field
+  // is:
+  // "command": ["git", "clone", "https://github.com/in-toto/demo-project.git"]
+  repeated string command = 1 [json_name="command"];
+
+  // Materials are the supply chain artifacts that go into the step and are used
+  // for the operation performed. The key of the map is the path of the artifact
+  // and the structure contains the recorded hash information. An example is:
```
aysylu: re "structure contains the recorded hash information": would it make sense to define it here as well then? If it's only ever going to be a hash, we have ways to represent this.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  //     "sha256": "ebebf..."
+  //   }
+  // }
+  map<string, google.protobuf.Struct> materials = 2 [json_name="materials"];
+
+  // Products are the supply chain artifacts generated as a result of the step.
+  // The structure is identical to that of materials.
+  map<string, google.protobuf.Struct> products = 3 [json_name="products"];
+
+  // By Products are made up of the data that is generated as part of a software
```
aysylu: nit: s/made up of the/

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+
+  // This is a field that can be used to capture information about the
+  // environment. It is recommended for this field to contain information that
+  // details environment variables, filesystem information, and the present
+  // working directory. The structure of this field is:
+  // "environment": {
+  //   "variables": "<ENV>",
+  //   "filesystem": "<FS>",
+  //   "workdir": "<CWD>"
+  // }
```
aysylu: Is this structure always expected to be the same in this context? If so, it'd be great to define it as a protobuf message as well.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  map<string, google.protobuf.Struct> environment = 5 [json_name="environment"];
```
aysylu: nit: here and everywhere, if `json_name` is the same as the field name, we don't need to include it, right?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// This corresponds to a full signed attestation - it is made up of one or more
```
aysylu: nit: full signed *in-toto* attestation -- to prevent confusion

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+// signatures and the core link attestation itself. This is used for occurrences
```
aysylu: is "core link attestation" a specific term in in-toto? if so, pointing to a definition of it, or having a few words to describe it at high-level would be helpful to the reader

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+
+// This file defines the proto that matches the schema of an in-toto link
+// file. in-toto is a framework to protect a software supply chain. You can read
+// more about in-toto at https://in-toto.io/ and you can find the current
+// version of the in-toto specification at
+// https://github.com/in-toto/docs/blob/master/in-toto-spec.md
+
+
+// A signature object consists of the KeyID used and the signature itself.
+message Signature {
```
aysylu: I thought I asked this during last iteration of the review, but can't find the discussion thread, so maybe not: could we utilize the `Signature` already defined in `common.proto`?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  // }
+  map<string, google.protobuf.Struct> environment = 5 [json_name="environment"];
+}
+
+// This corresponds to a full signed attestation - it is made up of one or more
+// signatures and the core link attestation itself. This is used for occurrences
+// of a Grafeas in-toto note.
+message Details {
+  repeated Signature signatures = 1 [json_name="signatures"];
+  Link link = 2 [json_name="signed"];
```
aysylu: So, a `Link` is an occurrence? and also a note?

Why are there multiple signatures and only 1 link?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// This contains the private and public portions of a key. The private portion
+// must be empty for a public key.
+message KeyValue {
```
aysylu: Could we utilize `Signature` in `common.proto` for this?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  string private = 1;
```
aysylu: What's the situation in which you'll have both `private` and `public` fields set? Or is it one or the other, but not both?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  string public = 2;
+}
+
+// This defines the format used to record keys used in the software supply
+// chain. An in-toto link is attested by a functionary using one or more keys
```
aysylu: A link to the `functionary` definition would be helpful to the reader here.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+// defined in the in-toto layout.
+message FunctionaryKey {
```
aysylu: It seems like there should be some overlap with the `Signature` proto in `common.proto` here. Is it missing something you need to represent in `FunctionaryKey`?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  // keyid is an identifier for the signing key.
+  string keyid = 1;
```
aysylu: s/keyid/key_id to be consistent with proto field naming, ditto for keytype and below

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+
+  // This field contains the actual private and public portions of the key.
+  KeyValue keyval = 4;
+
+  // This field contains the overall scheme used for the creation of the signing
+  // key. Eg: rsassa-pss-sha256
+  string scheme = 5;
+}
+
+// This contains the core fields corresponding to a software supply chain step.
```
aysylu: Does "core fields" have a specific meaning in in-toto parlance?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+// This information goes into a Grafeas note.
+message InToto {
```
aysylu: nit: let's move the note definition to the top of the file.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  string step_name = 1;
+
+  // This field contains the public keys that can be used to verify the
+  // signatures on the step metadata.
+  repeated FunctionaryKey signing_keys = 2;
+
+  // The following fields contain in-toto artifact rules identifying the
+  // artifacts that enter this supply chain step, and exit the supply chain
+  // step, i.e. materials and products of the step.
+  repeated google.protobuf.ListValue expected_materials = 3;
```
aysylu: you're already using `repeated` here, no need for `ListValue` IIUC

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+
+// This corrresponds to the core in-toto link attestation.
+message Link {
+  // This field contains the full command executed for the step. This can also
+  // be empty if attestations are generated for operations that aren't directly
+  // mapped to a specific command. Each term in the command is an independent
+  // string in the list. An example of a command in the in-toto metadata field
+  // is:
+  // "command": ["git", "clone", "https://github.com/in-toto/demo-project.git"]
+  repeated string command = 1 [json_name="command"];
```
aysylu: If the InToto note already lists the expected command, what's the rationale behind also including it in the occurrence?

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,145 @@
...
+  // step, i.e. materials and products of the step.
+  repeated google.protobuf.ListValue expected_materials = 3;
+  repeated google.protobuf.ListValue expected_products = 4;
+
+  // This field contains the expected command used to perform the step.
+  repeated string expected_command = 5;
+
+  // This field contains a value that indicates the minimum number of keys that
+  // need to be used to sign the step's link metadata.
+  int32 threshold = 6;
```
aysylu: any specific reason `int32` was picked over `int64`?

proto/v1beta1/intoto.proto
adityasaky: I'm going to briefly include some in-toto semantics here. in-toto has two types of metadata - [layouts](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#43-file-formats-layout) and [links](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#44-file-formats-namekeyid-prefixlink). A single layout is created to define the entire software supply chain, with all the keys used, the [steps](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#431-steps) that make up all the different operations, and how they're 'chained' together. Links are generated when these steps are actually performed.

Now, the thinking is to have a Grafeas note that corresponds to a step in the supply chain - it identifies the step by name, the keys to be used to verify the step's link metadata, [artifact rules](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#433-artifact-rules) that identify the files that were involved in the step (which are the materials and products), and an expected command which is the command that is to be used to perform this step (this can be blank). All these fields actually live in the layout of the supply chain, under the corresponding step. Every time the step is performed, we have link metadata being generated that is recorded as an occurrence.

> Why are there multiple signatures and only 1 link?

Let me verify something here.

proto/v1beta1/intoto.proto
adityasaky: These fields are set up as two-dimensional arrays of strings, this was the only way I could find to define in protobuf. Is there a better way? :)

A snippet from one of our demos:
```
   "expected_materials": [
     [
      "MATCH",
      "demo-project.tar.gz",
      "WITH",
      "PRODUCTS",
      "FROM",
      "package"
     ],
     [
      "ALLOW",
      ".keep"
     ],
     [
      "ALLOW",
      "alice.pub"
     ],
     [
      "ALLOW",
      "root.layout"
     ],
     [
      "DISALLOW",
      "*"
     ]
    ],
```

proto/v1beta1/intoto.proto
adityasaky: The occurrence includes the actual command run - in-toto verification compares the two.

proto/v1beta1/intoto.proto
adityasaky: The structure mentioned here is the suggested format - the environment field may be populated to hold other relevant information necessary for a given context. I should update the comment to indicate as much.

proto/v1beta1/intoto.proto
adityasaky: It does not, I see that this has come up a few times and I can see why it can be confusing. I'll update the comments here and elsewhere to be clearer.

proto/v1beta1/intoto.proto
adityasaky: No, I'll update it to use `int64`.

proto/v1beta1/intoto.proto
adityasaky: > Why are there multiple signatures and only 1 link?

The in-toto metaformat allows for multiple signatures because theoretically a link can have more than 1 signature. The metaformat also carries over to layouts which usually do contain 1+ signatures. While the links actually passed to the Grafeas may only contain a single signature, we need `repeated` to match the in-toto metaformat.

proto/v1beta1/intoto.proto
adityasaky: I initially defined the general key format we use where `private` is left empty to declare public keys. I'm going to re-work this to drop the private field as in this context the keys are always public keys.

proto/v1beta1/intoto.proto
adityasaky: Dropping `private` resolves this.

proto/v1beta1/intoto.proto
adityasaky: Note: I'm renaming `FunctionaryKey` to `SigningKey` to avoid confusions with the term Functionary.

As for overlap, as I understand, `Signature` defines a specific signature made using a particular key. `SigningKey` is a definition of such a key which can then be used to verify signatures. So I think they need to be distinct.

proto/v1beta1/intoto.proto
adityasaky: We could use the format specified there for the key of a materials/products object (however this will require ITE-4: https://github.com/in-toto/ITE/pull/6/files#diff-669c59f42864911b31bda53fe6b3d41e) but the hash object allows us to support non SHA-256 hashes as well as multiple hash algorithms for a particular artifact.

proto/v1beta1/intoto.proto
adityasaky: We don't have a strict structure for it as we don't enforce specific hash algorithms. It can also be multiple as an artifact can have its hashes recorded using more than one algorithm.

adityasaky: Also, changes here **_could_** mean that when it's serialised into a JSON from the Grafeas server, it would no longer match the in-toto format and signature verification will fail.

proto/v1beta1/intoto.proto
adityasaky: Looking at the `Signature` format defined in `common.proto`, we noticed a couple of things. For one thing, our Key IDs are not RFC3986 complaint - they're just hash strings that refer to a specific key definition (which here happens with `SigningKey`). Secondly, one of the examples is a reference to RFC6920, which is still proposed.

proto/v1beta1/intoto.proto
SantiagoTorres: I think it's wortwhile keeping the hash-objects, and then we can massage both implementations later down the line to avoid redundancy. I'm curious to know if there's a way to define hash values in such a way that you can leverage multiple-hashing algorithms at once. 

As an aside, some of these definitions (e.g., `pip`) may could be improved to include namespacing and resource-location. I don't think the current schema supports uses like self-hosted pypi registries and more. In the end, pip is just a frontend to the pypi/warehouse/other package repositories...

docs/grafeas_concepts.md
adityasaky: Added a link to the specification.

proto/v1beta1/intoto.proto
aysylu: Gotcha, so here you'll want to have a `repeated Material expected_materials = 3;` field, and define a `Material` proto message separately. If there's no structure to `Material` beyond having a list of strings, then you'd just have 1 field in the proto, `repeated string ... = 1;`. If you do know the structure ahead of time, then you'd define it in the `Material` proto itself. Does this make sense?

proto/v1beta1/intoto.proto
aysylu: Makes sense. Let's rename this to `effective_command` or `command` vs `expected_command` in the note, or something along this lines? This way, the relationship between these will be obvious to the reader.

docs/grafeas_concepts.md
aysylu: Great, thank you!

proto/v1beta1/intoto.proto
adityasaky: I tried that before settling for `google.protobuf.ListValue` and the issue I ran into IIRC is that it expected a list of an object (which could handle repeated strings for the given key).

```
[
  {
    <key as per the proto definition>: [<list of strings>]
  }
]
```

I can double check this though.

proto/v1beta1/intoto.proto
adityasaky: Renamed to `effective_command`. Note: we'll need `json_name` for this one as it's `command` in the in-toto metadata and here the in-toto link is embedded directly, unlike in the note.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+package grafeas.v1beta1.intoto;
+
+message SignatureObject {
```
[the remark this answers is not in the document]
adityasaky: This point has come up in a thread below, resolving this instance.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,135 @@
...
+
+  // This field contains the public keys that can be used to verify the
+  // signatures on the step metadata.
+  repeated SigningKey signing_keys = 2;
+
+  // The following fields contain in-toto artifact rules identifying the
+  // artifacts that enter this supply chain step, and exit the supply chain
+  // step, i.e. materials and products of the step.
+  repeated google.protobuf.ListValue expected_materials = 3;
+  repeated google.protobuf.ListValue expected_products = 4;
```
aysylu: The previous thread on this is locked for some reason, so continuing the conversation here, [context](https://github.com/grafeas/grafeas/pull/391#discussion_r369306952):

So to make sure we're talking about the same thing, here's my suggestion:

```
message Material {
  repeated string value = 1; // there's probably a better nam than `value` for this, so I defer to you
}

message InToto {
  ...
 repeated Material expected_materials = 3;
 repeated Material expected_products = 4;
}
}
```

You could also next the `Material` definition [inside the InToto proto](https://developers.google.com/protocol-buffers/docs/proto3#nested), right before the field.

I'm guessing the main concern here is how it's generated in JSON, correct? How would it be represented with the proposal above vs what you need it to look like?

proto/v1beta1/intoto.proto
aysylu: Makes sense, thanks for clarifying!

proto/v1beta1/intoto.proto
aysylu: I'm confused. What I'm wondering is what does `foo/bar` represent here? It sounds like it's supposed to represent an artifact, based on `The key of the map is the path of the artifact`. If so, we represent all artifacts using CPE URIs, and I'd like to keep this consistent. In Grafeas, the artifacts are binaries, packages, containers, etc. Do we have the same definition here?

```
"foo/bar": {
  "sha256": "ebebf..."
   }
}
```

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,135 @@
...
+
+  // Materials are the supply chain artifacts that go into the step and are used
+  // for the operation performed. The key of the map is the path of the artifact
+  // and the structure contains the recorded hash information. An example is:
+  // "materials": {
+  //   "foo/bar": {
+  //     "sha256": "ebebf..."
+  //   }
+  // }
+  map<string, google.protobuf.Struct> materials = 2;
```
aysylu: Here, you'd want to re-use the same definition of `Material` as suggested above, then instead:

```
map<string, Material> materials = 2;
```

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,135 @@
...
+
+  // Products are the supply chain artifacts generated as a result of the step.
+  // The structure is identical to that of materials.
+  map<string, google.protobuf.Struct> products = 3;
```
aysylu: ditto here for the `Material`; you could also define another type, `Product`, in case you ever expect the definitions to diverge.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,135 @@
...
+
+  // By Products are data generated as part of a software supply chain step, but
+  // are not the actual result of the step. The suggested fields are "stderr",
+  // "stdout", and "return-value".
+  map<string, google.protobuf.Struct> byproducts = 4;
```
aysylu: ditto here for defining a separate type

In general, defining specific message types is a very common practice in proto design.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,135 @@
...
+  // environment. It is suggested for this field to contain information that
+  // details environment variables, filesystem information, and the present
+  // working directory. The recommended structure of this field is:
+  // "environment": {
+  //   "variables": "<ENV>",
+  //   "filesystem": "<FS>",
+  //   "workdir": "<CWD>",
+  //   "<ANY OTHER RELEVANT FIELDS>": "..."
+  // }
+  map<string, google.protobuf.Struct> environment = 5;
```
aysylu: ditto here for the specialized `Environment` proto message. The fields are what you already listed: `variables`, `filesystem`, `workdir` etc. And you can always add more fields to the proto message without having to release new API. However, since this is `v1beta1` API version, some of this could be ok, assuming all dependent software is updated accordingly.

proto/v1beta1/intoto.proto
adityasaky: `byproducts` and `environment` have no strict structure - they can be empty - and their contents are not verified by in-toto. For example, here's an example in-toto link that Datadog is using: [tag.c295cf63.link](https://dd-integrations-core-wheels-build-stable.datadoghq.com/targets/in-toto-metadata/3e592d47ab815cb9bd8aa064bbaa2a217c1baa9a55a7f8b73d2da6dfd14a247e/tag.c295cf63.link). They don't populate either field.

proto/v1beta1/intoto.proto
adityasaky: as above, those fields are just some examples we came up with, but they may not occur in practice.

proto/v1beta1/intoto.proto
adityasaky: Yes. I built this locally and the metadata needs to look like:
```
        "expectedProducts": [
          {
            "value": [
              "CREATE",
              "demo-project/foo.py"
            ]
          },
          {
            "value": [
              "DISALLOW",
              "*"
            ]
          }
        ],
```
instead of:
```
        "expectedProducts": [
          [
            "CREATE",
            "demo-project/foo.py"
          ],
          [
            "DISALLOW",
            "*"
          ]
        ],
```

This is what led me to use `ListValue` to allow for nested lists _without_ an intermediate map, so to speak.

proto/v1beta1/intoto.proto
aysylu: Hm, from the description above, the structure is `artifact: values`, where `artifact` is a binary/pkg/container in Grafeas parlance, so is represented by `cpe_uri` -- am I misunderstanding the description in the comment? If that's the case, how about making just a generic

```
message Material {
  map<string, string> artifact_to_materials = 1;
}
```
?

Also, PTAL at the proto design resources I shared, it might be helpful in clarifying what I'm asking here.:)

proto/v1beta1/intoto.proto
aysylu: Got it! In this case, PTAL at my suggestion from a few minutes ago about structuring the `Material` proto message with a map as the field.

proto/v1beta1/intoto.proto
adityasaky: I think we may be going around in circles a bit here, and I may be misunderstanding your suggestion. So let me summarise for my own clarity. The `materials` and `products` field in link metadata are maps where each key is an artifact identifier (currently file paths), and each value is another map. The keys in this inner map are hash algorithms, and the values are the hashes of the artifact using the corresponding hash algorithm.
```
"materials":  {
  "<artifact>: {
    "sha256": "<hash>",
    "sha512": "<hash>",
  }
}
```
What I'm really trying to define using protobuf is `map<string, map<string, string>> materials = 2;`. When I found I couldn't do that, the closest I found in the docs was [google.protobuf.String](https://developers.google.com/protocol-buffers/docs/proto3#json) in place of the inner map. I'd like to make this specific, and ensure that the inner map also just contains `string: string` but I haven't found a way so far. I did try using a message as follows:
```
  message ArtifactHashes {
    map<string, string> hashes = 1;
  }

  map<string, ArtifactHashes> materials = 2;
```
This has the effect of requiring `hashes` as a key in the JSON we POST to the server, which differs from the in-toto spec. Example:
```
          "materials": {},
          "products": {
            "demo-project/foo.py": {
              "hashes": {
                "sha256": "ebebf8778035e0e842a4f1aeb92a601be8ea8e621195f3b972316c60c9e12235"
              }
            }
          },
```

proto/v1beta1/grafeas.proto
```diff
@@ -34,6 +34,7 @@ import "proto/v1beta1/image.proto";
 import "proto/v1beta1/package.proto";
 import "proto/v1beta1/provenance.proto";
 import "proto/v1beta1/vulnerability.proto";
+import "proto/v1beta1/intoto.proto";
```
wkozlik: A good convention to follow is to sort the import statements.
See https://developers.google.com/protocol-buffers/docs/style

adityasaky: Fixed. :)

proto/v1beta1/intoto.proto
adityasaky: Resolving this as we're discussing this in a separate thread below.

proto/v1beta1/intoto.proto
adityasaky: `foo/bar` represents a path in the file system to individual components that make up a software package as they move through the supply chain. This is going to be more generalised with the work in [ITE-4](https://github.com/in-toto/ITE/pull/6) which allows for URIs that refer to more abstract artifacts.

As I understand the CPE specification, I'm not seeing a lot of overlap between our artifact URIs and CPE URIs - they don't represent the components within the package's supply chain but rather the package as a whole. Am I misunderstanding the spec? Please let me know :)

proto/v1beta1/intoto.proto
wkozlik: Is the list of supported hash algorithms known? If so, we could define the `ArtifactHashes` as:
```
  message ArtifactHashes {
    string sha256 = 1;
    string sha512 = 2;
  }
```

proto/v1beta1/intoto.proto
aysylu: Ok, great, there *is* a way to represent files in Grafeas as well, please see [resource urls](https://github.com/grafeas/grafeas/blob/master/docs/grafeas_concepts.md#resource-urls), `Generic file`.

Yes, the idea here is that an artifact in Grafeas is a file, and a package in which the file is a component, and the container that might contain packages. Here's an example of tracking software supply chain (SSC) for an individual file:

file (eg source code, test, data etc) -> built into a package -> tested -> deployed.

So we can track the lifetime of a file through SSC if we need to, it can even be at finer granularity if we need it, eg block of code, etc. CPE specification is just a generic way to represent this. Does this make sense?

proto/v1beta1/intoto.proto
adityasaky: The specification does not enforce any specific hash algorithms. The in-toto reference implementation currently implements SHA-256. I'm okay to use this definition and we can approach this further down the line, as and when the algorithms need to be updated. :)

adityasaky: Added.

proto/v1beta1/intoto.proto
aysylu: Resolving this to condense the various threads about the same thing.

proto/v1beta1/intoto.proto
aysylu: merging into 1 thread, resolving here

proto/v1beta1/intoto.proto
aysylu: merging into 1 thread

proto/v1beta1/intoto.proto
aysylu: merging into 1 thread :)

proto/v1beta1/intoto.proto
adityasaky: I think the confusion is that in in-toto, the `foo/bar` points to a map with hash values for `foo/bar`. The `Generic File` URI contains very similar information except it includes everything in a single string. The difference is that the map allows us to record more than one hash, using more than one algorithm simultaneously.

Are you suggesting we update `foo/bar` to the `Generic File` URI format and have it point to the map with the hash values? Something like:

```
"file://sha256:244fd47e07d1004f0aed9c156aa09083c82bf8944eceb67c946ff7430510a77b:foo/bar" {
  "sha256": "244fd47e07d1004f0aed9c156aa09083c82bf8944eceb67c946ff7430510a77b",
}
```

aysylu:
Gathering the couple threads on the same question of representation of in-toto metadata in Grafeas protos:

@adityasaky this is my proposal to represent the materials information in the in-toto link, including yours and @wkozlik's from the discussion:

```
message ArtifactHash {
    string sha256 = 1;
}

message Material {
  Resource resource = 1;
  repeated ArtifactHash = 2;
}

message Intoto {
  ...
  repeated Material expected_materials = 3;
  repeated Material expected_products = 4;
}

message Link {
  ...
  repeated Material materials = 2;
  repeated Material products = 3;
}
```

defined in `grafeas.proto`, used here: [Resource](https://github.com/grafeas/grafeas/blob/ffb2520179bdf591987e57cd2ba9b0bbaa053494/proto/v1beta1/grafeas.proto#L251).

For Hashes representation, you have a couple additional options:
1. Use [this representation](https://github.com/grafeas/grafeas/blob/404aefd09564738e4c1a238c13d50567455a39f5/proto/v1/common.proto#L102) , which includes the algorithm in its path, is already in `common.proto` so can be imported directly. To make this work, you may need additional processing to get into the JSON format you're looking for.
2. Extract [this representation](https://github.com/grafeas/grafeas/blob/2966591cc0a5a8f02f28fc8d0584e305adb9500f/proto/v1beta1/provenance.proto#L111) or just the `enum` into `common.proto` and import in `intoto.proto`. You'd probably want to add `SHA512` field to the `enum` or the `message`, which is reasonable. This assumes that an artifact will only have 1 type of hash, but it sounds like you'd like several hashes for the same artifact.

The reason I'm suggesting the `enum` in the first place is that it sounds like an artifact will have only 1 hash associated with it? If that's not the case, then I agree that the `ArtifactHashObject` is a good way to go for this, but I'd recommend renaming this to `ArtifactHash` for consistency with proto design in this project.

To confirm that the proposed representation works, from the [example in in-toto documentation](https://in-toto.io/examples/debian.html):

```
// products field in Link proto is repeated Material field
// |
// v
"products": {
      "/home/lukas/demo/grep_2.12-2.debian.tar.bz2": {  // <- key is a Resource field
        "sha256": "37887d8aecec66e75365abd8c41be94f75e7a3acf1d6b27fc121584f47281525"
        "sha512": "beef",  // <- value is a repeated ArtifactHash
      },
      "/home/lukas/demo/grep_2.12-2.dsc": {
        "sha256": "a6a63fd21da40d11ce6ae2bc6f633bd27cd206c2348b2ef306e1b68120f7e58e"
      },
      "/home/lukas/demo/grep_2.12.orig.tar.bz2": {
        "sha256": "0119987171cd60b87c89557524fc6636bdad770dae71917adcaef6abffb1be67"
      }
}
```

WDYT?

proto/v1beta1/intoto.proto
aysylu: It could be `"file://sha256:244fd47e07d1004f0aed9c156aa09083c82bf8944eceb67c946ff7430510a77b:foo/bar"` and have it build in the hash inside, or `"file://foo/bar"` etc.

SantiagoTorres:
Hmm, I'm quite struggling to find a rationale to embedd everything in lists. From my experience this approach has just made looking up resource identifiers `O(n)` instead of `O(1)`. I remember in earlier conversations with the Grafeas team one of the goals was to have efficient query-ability, is that not the rationale anymore, @aysylu ?

aysylu:
> Hmm, I'm quite struggling to find a rationale to embedd everything in lists. From my experience this approach has just made looking up resource identifiers `O(n)` instead of `O(1)`. I remember in earlier conversations with the Grafeas team one of the goals was to have efficient query-ability, is that not the rationale anymore, @aysylu ?

Couple clarifications:

1. What do you mean by embedding everything in lists? My proposal above is just a way things are represented in protos. `ListValue`, `Struct` etc aren't used in proto design, these are how conversion to JSON happens.
2. What do you mean by having efficient queryability? Once the Grafeas <-> in-toto conversion happens, we get benefits from both of the systems, with no performance penalty. You can continue speaking the Grafeas format to query the Grafeas server.
3. Also, I don't remember having a discussion about making this conversion as efficient as possible. Could you please give more context on this?

SantiagoTorres:
> 1. What do you mean by embedding everything in lists? My proposal above is just a way things are represented in protos. ListValue, Struct etc aren't used in proto design, these are how conversion to JSON happens.

Sure, I don't think I can claim I know this to be otherwise. I think you know better than me what can and can't be done.

> 2. What do you mean by having efficient queryability? Once the Grafeas <-> in-toto conversion happens, we get benefits from both of the systems, with no performance penalty. You can continue speaking the Grafeas format to query the Grafeas server.

My understanding is that, from my original discussions with the team, some decisions as to how the protos look were made so it was easy to query which Notes/Ocurrences had certain characteristics. What I got back then is that the idea was that approaches such as "just stick it all in a string" or "iterate over all of the elements on a list to see if this list contains element X" would be very expensive when doing these types of queries. I think that using the format suggested will incur in those penalties. 

At the end of the day, you know better than I do about what the goals of Grafeas are. I'm just curious to know if that perspective changed (it can happen, it's been 3 years since).

> 3 Also, I don't remember having a discussion about making this conversion as efficient as possible. Could you please give more context on this?

I didn't mean to say that. The conversion can definitely happen, and I assume the requirement of it being efficient or not will start to fall on whether users are affected by performance.

Sorry if I wasn't clear or gave the wrong impression

wkozlik:
Maybe we could take a step back to make sure we are on the same page about high-level goals and assumptions?

Feel free to correct my statements below if anything is unclear or inaccurate.

Goals
1) to store in-toto metadata in Grafeas
2) in-toto protobuf definitions should be consistent with the rest of Grafeas protos
3) in-toto protobuf definitions should be identical or very close to existing in-toto JSON schema

Because achieving goals (2) and (3) at the same time can be difficult, we need to find a middle ground.

Assumptions
* in-toto layout will not be stored in Grafeas
* in-toto step will be stored in Grafeas as a Note
* in-toto link will be stored in Grafeas as an Occurrence
* a new Note kind "INTOTO" will be added

Questions:
1) Where is the in-toto layout stored?

2) Is the in-toto layout immutable? If so, is the in-toto Grafeas Note also immutable?

3) Is the in-toto link immutable?

4) An occurrence has a field `resource`. What does this field mean in the context of an in-toto link?

```
{
  "name": "projects/intoto-project/occurrences/note_id",
  "resource": {
    "uri": "https://gcr.io/project/image@sha256:foo"
  },
  "noteName": "projects/intoto-project/notes/note_id",
  "kind": "INTOTO",
  "intoto": {
    "..."
  }
}
```

5) What are potential problems with in-toto protobuf definitions being different than the existing in-toto JSON schema?

6) Would those problems make the adoption of in-toto with Grafeas more difficult?

7) Some in-toto metadata fields have no strict structure. Does it mean that users decide what fields to put in those structures?

adityasaky:
I'm going to take a stab at first answering @wkozlik's questions.

I think the goals you've listed are accurate, and you're right that achieving 2 and 3 simultaneously is a little complicated.

Assumptions:

Your assumptions are also pretty much on point. The new Note kind corresponds to a single step instance in an in-toto layout, and the occurrences for each one of these is an actual link. So the note tells us what to expect from its occurrences.

Questions:

1. The root layout would probably be stored in an admissions controller.

2. All in-toto metadata is signed, so technically they are immutable? But they can be and are sometimes replaced. This of course needs them to be re-signed.

3. Pretty much the same as above.

4. This would probably correspond to the `Resource` whose supply chain these links are a part of, unless I'm misunderstanding the field.

5. It'd mean that default in-toto links can't be pushed to a Grafeas server. For example, the in-toto reference implementation's `runlib`, which generates links, automatically signs them. For this compatibility, we'll need a plugin that intervenes before the signing, modifies the structure of the link recorded, and then sign it before pushing to a Grafeas instance. Similarly, a link fetched from a Grafeas server for in-toto verification will need to have its format changed back after signature validation, and before the rest of the verification workflow.

6. They theoretically could - third party implementations would have to also handle the conversion in some manner.

7. Yes, they're left for implementors/users to record any context relevant information that they wish to use for verification using inspections. In fact, we need identify how to incorporate these without using `Struct` as by nature they have no definition.


--------------

@aysylu

I think Santiago meant that by using `repeated Material`, we have a list of objects that each correspond to a single artifact and its hash(es), meaning we can't look up the hash information for an artifact in O(1). However, if we're going with the conversion from the Grafeas-like metadata scheme back to the default in-toto scheme pre-verification, I think it won't affect the efficiency of verification itself, but we'd have to account for the conversion. But implementing the conversion shouldn't be problematic.

Now to your suggestion from the previous comment: I think they'll work with the proposed conversion in between for the `Materials` and using the `Resource.uri` to identify them - the conversion would handle taking the artifact identifier (the file path in the examples) and build from the `Generic File` by including the scheme and hash. And an artifact can have more than hash simultaneously recorded for it using different algorithms, so I agree we can retain `ArtifacthashObject` with the proposed renaming.

The suggestion for `expected_materials` and `expected_products` don't quite work as their formats differ from the `Material` field - they are nested lists. I think the conversion should go from:
```
"expected_products": [["CREATE", "<something>"], [<ANOTHER RULE>]],
```
to:
```
"expected_products": [{"artifact_rule": ["CREATE", "<something>"]}, {"artifact_rule": [<ANOTHER RULE]}]
```
This is what defining a separate `message` for artifact rules gets us, and I think it's more aligned with the Grafeas proto definitions.

That leaves us with defining `environment` and `byproducts` without using `Struct`, and nailing down the specifics of how we handle this conversion in practice.

I'll push up changes for artifacts and artifact rules, while we nail down `environment` and `byproducts`. :)

adityasaky:
@aysylu @wkozlik 

In order to use `Resource`, I think I have to move it to `common.proto` from `grafeas.proto` and then import `common.proto` in `intoto.proto`? I can't import `grafeas.proto` directly because of circular imports. Does that sound right?

wkozlik:
I am not 100% sure, but I am afraid that moving the `Resource` message from `grafeas.proto` to `common.proto` would be a breaking change for existing users. I wonder if the `Material` message with explicit `resource_uri` would be acceptable. 
```
message Material {
  string resource_uri = 1;
  repeated ArtifactHash = 2;
}
```

wkozlik:
Regarding the fields `byproducts` and `environment`, would it work to have the following definition for them?
```
message Byproducts {
  map<string, string> values = 1;
}

message Environment {
  map<string, string> values = 1;
}

message Link {
  ...
  Byproducts byproducts = 4;
  Environment environment = 5;
```

adityasaky:
@wkozlik I do see `string resource_uri` in `deployment.proto` so I think I'll go with that for now, thanks! As for your other suggestion, can I use `map<string, string>` directly in `Link` to save the addition of the `values` field to the metadata? 

```message Link {
  ...
  map<string, string> byproducts = 4;
  map<string, string> environment = 5;
```

The conversion process will string-ify all content and that should take care of that. Thoughts?

wkozlik:
At this point we will not avoid the conversion between the JSON schema and the protobuf definition. We could try to simplify this conversion here and there, but having specific type for `Environment` and `Byproducts` doesn't make much difference from that perspective. However, having those types makes the API a little more extensible in the future. My thought was that if in the future we would want to add explicit fields to either of these types, we would be ready for it. But with generic types like a `struct` or a `map` it would be a bigger change. 

In general, I feel like protobuf was designed with backward compatibility in mind, and having that in mind could help us in a long run.

adityasaky:
Ah I see what you mean. I pushed up the change. Also, I've opted for `custom_values` in place of `values` so that it better indicates it's for custom fields as opposed to any fields we may define in future. I also updated some of the comments to indicate the new structures. :)

adityasaky:
I think all the points of concerns from the last iteration or two of reviews have been addressed? Let me know if things look okay now, @wkozlik @aysylu.

wkozlik:
One last question. Should this be `hash` instead of `hashes`?
```
message Artifact {
    string resource_uri = 1;
    ArtifactHash hashes = 2;
  }
```

adityasaky:
I've left that open for support for additional algorithms (an implementation may choose to record the hash with more than one algorithm simultaneously). As we have implementations support it, I'll be updating `ArtifactHash` with additional fields, for example `sha512` and so on. It occurs to me that perhaps the field should be `ArtifactHashes`. What do you think?

wkozlik:
I agree, if multiple hashes will be supported, then the type name `ArtifactHashes` makes sense.

adityasaky:
Done.

wkozlik:
Okay, great. Let me pull this PR locally and run a few tests to confirm all is working as expected.

wkozlik:
LGTM

wkozlik:
This is an example of an in-toto occurrence that I tested with:
```json
{
    "name": "projects/occurrence_example/occurrences/intoto-test",
    "resource": {
        "uri": "https://gcr.io/project/image@sha256:foo"
    },
    "noteName": "projects/provider_example/notes/testNote",
    "kind": "INTOTO",
    "intoto": {
        "signatures": [
            {
                "keyid": "12c55e46654c682d3ffd3b63492adf422e6812eb1ac41574d083b9e770d1e4c2",
                "sig": "a0d3237a59a96343b2e2caee2bf0e27a249a01f796c3fec53f184fc2720b44dd7b2..."
            }
        ],
        "signed": {
            "command": [
                "dget",
                "http://cdn.debian.net/debian/pool/main/g/grep/grep_2.12-2.dsc"
            ],
            "materials": [],
            "products": [
                {
                    "resourceUri": "/home/lukas/demo/grep_2.12-2.debian.tar.bz2",
                    "hashes": {
                        "sha256": "37887d8aecec66e75365abd8c41be94f75e7a3acf1d6b27fc121584f47281525"
                    }
                },
                {
                    "resourceUri": "/home/lukas/demo/grep_2.12-2.dsc",
                    "hashes": {
                        "sha256": "a6a63fd21da40d11ce6ae2bc6f633bd27cd206c2348b2ef306e1b68120f7e58e"
                    }
                },
                {
                    "resourceUri": "/home/lukas/demo/grep_2.12.orig.tar.bz2",
                    "hashes": {
                        "sha256": "0119987171cd60b87c89557524fc6636bdad770dae71917adcaef6abffb1be67"
                    }
                }
            ],
            "byproducts": {
                "customValues": {
                    "return_value": "1",
                    "stderr": "--2017-08-08 11:05:06--  http://cdn.debian.net/debian/pool/main/g/g...",
                    "stdout": "dget: retrieving http://cdn.debian.net/debian/pool/main/g/grep/grep..."
                }
            },
            "environment": {
                "customValues": {}
            }
        }
    }
}
```

wkozlik:
Note: to fix the ci/circleci build error you may need to merge master branch into your `adityasaky:add-in-toto-proto` branch.

SantiagoTorres:
I wonder if for the sake of having a cleaner history @adityasaky could rebase on master and then push?

wkozlik:
I believe we still need to do a pull request.  For synching the fork this info is useful: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork?

SantiagoTorres:
no, another pull request won't be necessary. What I'm talking about is:

1. `git pull --rebase upstream master` (with `--rebase` so as to not generate another merge commit, given that this won't be a FF merge) <- this is where we differ
2. Push to this branch to update the remote
3. the PR is updated and there shouldn't be more conflicts.

adityasaky:
There we go, let me know if this looks okay, @wkozlik. :)

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+  // "materials": [
+  //   {
+  //     "resource_uri": "foo/bar",
+  //     "hashes": {
+  //       "sha256": "ebebf...",
+  //       <OTHER HASH ALGORITHMS>: <HASH VALUE>
+  //     }
+  //   }
+  // ]
+  repeated Artifact materials = 2;
```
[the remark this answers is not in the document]
adityasaky: No, `expected_materials` contains artifact rules that define what materials are expected in a particular step. `materials` record what were actually used as materials and their hashes. The same goes for `expected_products`.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+
+// This defines the format used to record keys used in the software supply
+// chain. An in-toto link is attested using one or more keys defined in the
+// in-toto layout.
+message SigningKey {
+  // key_id is an identifier for the signing key.
+  string key_id = 1;
+
+  // This field identifies the hash algorithms used in the particular key ID.
+  repeated string key_id_hash_algorithms = 2;
```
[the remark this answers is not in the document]
adityasaky: No, but I think this can be handled in the conversion to and from the in-toto metadata format, so I think I'll get rid of it.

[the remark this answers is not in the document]
adityasaky: The algorithms listed in this field can be used to verify the key ID but we can simplify the proto definition by handling it in the conversion layer. :)

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+// This corresponds to a signed in-toto link - it is made up of one or more
+// signatures and the in-toto link itself. This is used for occurrences of a
+// Grafeas in-toto note.
+message Details {
+  repeated Signature signatures = 1;
+  Link link = 2 [json_name="signed"];
+}
+
+// A signature object consists of the KeyID used and the signature itself.
+message Signature {
```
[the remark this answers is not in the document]
adityasaky: While the format matches the in-toto definition, I think it goes back to a couple of things:
1. in-toto key IDs are not RFC3986 compliant - they're just hashes.
2. Further, they're also not RFC6920 compliant - while this can probably be handled in the conversion layer, there's also the matter that it's still a proposal.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+  // This field contains a value that indicates the minimum number of keys that
+  // need to be used to sign the step's in-toto link.
+  int64 threshold = 6;
+}
+
+// This defines the format used to record keys used in the software supply
+// chain. An in-toto link is attested using one or more keys defined in the
+// in-toto layout.
+message SigningKey {
+  // key_id is an identifier for the signing key.
```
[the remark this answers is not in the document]
adityasaky: I'm including an example.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+  // This field identifies the specific signing method. Eg: "RSASSA-PSS",
+  // "ed25519", and "ecdsa".
+  string key_type = 3;
+
+  // This field contains the actual public key.
+  string public_key_value = 4;
+
+  // This field contains the overall scheme used for the creation of the signing
+  // key. Eg: "rsassa-pss-sha256".
+  string key_scheme = 5;
```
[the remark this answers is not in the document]
adityasaky: `key_type` is a string denoting a public key signature system, such as "rsa" or "ecdsa". `key_scheme` is a string denoting a corresponding signature scheme.  For example: "rsassa-pss-sha256" and "ecdsa-sha2-nistp256".

I corrected an error in the example I'd given for `key_type`. I've removed the hash algorithms field as I mentioned in the other thread but I advocate for retaining the `key_type` and `key_scheme` fields.

proto/v1beta1/intoto.proto
```diff
@@ -0,0 +1,163 @@
...
+  // This field identifies the name of the step in the supply chain.
+  string step_name = 1;
+
+  // This field contains the public keys that can be used to verify the
+  // signatures on the step metadata.
+  repeated SigningKey signing_keys = 2;
+
+  // Defines an object to declare an in-toto artifact rule
+  message ArtifactRule {
+    repeated string artifact_rule = 1;
```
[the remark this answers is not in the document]
adityasaky: This field contains the rule itself in the form `['MATCH', <something>, 'FROM', <step name>]`. I think renaming to something else like just `rule` is actually less descriptive. What do you think?

proto/v1beta1/intoto.proto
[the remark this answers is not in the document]
adityasaky: Artifact rules are documented here: https://github.com/in-toto/docs/blob/master/in-toto-spec.md#433-artifact-rules

proto/v1beta1/intoto.proto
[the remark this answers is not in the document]
aysylu: Got it! I'll leave this up to you whether to rename it to rule or keep as is, either works.:)

proto/v1beta1/intoto.proto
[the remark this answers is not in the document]
aysylu: Got it, thanks for clarifying

proto/v1beta1/intoto.proto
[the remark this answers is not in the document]
aysylu: Makes sense, thanks

proto/v1beta1/intoto.proto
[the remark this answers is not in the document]
aysylu: Thanks for clarifying

aysylu:
Welcome to Grafeas, in-toto metadata! :D Thanks for all your hard work, @adityasaky, @wkozlik, and @SantiagoTorres!

adityasaky:
@aysylu I'm afraid I was waiting on a response on a couple of points before pushing up the last nits and corrections, and they've been left out. Shall I PR them right now separately?

aysylu:
@adityasaky, sorry about that! Yes, please, a separate PR would be great.

adityasaky:
Done, #426. Apologies!

Issue #389, opened 2019-10-11
The REST API is not working with the current code at master


judavi:
## Expected Behavior

The REST API should list the projects available using the REST endpoint /v1beta/projects

## Actual Behavior

The REST API is replying with an error:
```
{"error":"all SubConns are in TransientFailure, latest connection error: timed out waiting for server handshake","message":"all SubConns are in TransientFailure, latest connection error: timed out waiting for server handshake","code":14,"details":[]}%
```

## Steps to Reproduce the Problem

1.  Download the master branch
2.  Execute 
```
make build 
```
3.  Go to v1beta1/main
```
cd $GOPATH/src/github.com/grafeas/grafeas/go/v1beta1/main
```
4.  Create a config file for connecting with a local Postgres database
```
cat config.yaml
grafeas:
  api:
    address: "localhost:8080"
    cafile:
    keyfile:
    certfile:
    cors_allowed_origins:
  storage_type: "postgres"
  postgres:
    host: "0.0.0.0:5432"
    dbname: "postgres"
    user: "postgres"
    password: "password"
    sslmode: "disable"
    paginationkey:

```
5. Run main.go using the config file using GRPC variables to get more information 
```
GRPC_GO_LOG_VERBOSITY_LEVEL=99 GRPC_GO_LOG_SEVERITY_LEVEL=info GRPC_VERBOSITY=DEBUG go run main.go --config config.yaml
```
6.  Execute a GET call to the projects API
```
curl http://localhost:8080/v1beta1/projects
```

## Excepted output
```
{"projects":[],"nextPageToken":""}
```

## Current output client 
```
{"error":"all SubConns are in TransientFailure, latest connection error: timed out waiting for server handshake","message":"all SubConns are in TransientFailure, latest connection error: timed out waiting for server handshake","code":14,"details":[]}%
```

## Current output server
```
INFO: 2019/09/16 20:43:56 parsed scheme: ""
INFO: 2019/09/16 20:43:56 scheme "" not registered, fallback to default scheme
INFO: 2019/09/16 20:43:56 ccResolverWrapper: sending update to cc: {[{[::]:8080 0  <nil>}] <nil>}
INFO: 2019/09/16 20:43:56 ClientConn switching balancer to "pick_first"
2019/09/16 20:43:56 grpc server is configured without client certificate authentication
INFO: 2019/09/16 20:43:56 pickfirstBalancer: HandleSubConnStateChange: 0xc000032030, CONNECTING
[...]

PR #395, opened 2019-11-04
Rest endpoint fix 

judavi:
Hello @aysylu !
This will fix the issue from #389 :)

### DIFF
--- a/go/v1beta1/server/server.go
+++ b/go/v1beta1/server/server.go
@@ -122,8 +122,8 @@ func run(config *config.ServerConfig, db *grafeas.Storage, proj *project.Storage
 
 		log.Println("grpc server is configured with client certificate authentication")
 	} else {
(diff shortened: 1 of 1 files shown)


judavi:
Hi @aysylu ! I took inspiration from this answer https://github.com/grpc-ecosystem/grpc-gateway/issues/727#issuecomment-413715690  and one of the comments "Without tls all works fine with cmux, i can connect without certs to grpc server and do some rest requests, but with tls enabled I'm always have timeouts"

aysylu:
It looks like our code didn't change between `v0.1.0` and `v0.1.1` (just some copy/paste to refactor code). My suspicion is something in `cmux` changed between Aug 7 (the release of `v0.1.0`) and Aug 16 (the report of the issue @judavi linked to). I haven't dug into what changed during then, but @noahklein was able to confirm in his clone that the fix solves the issue. Merging. Thanks for contributing the fix, @judavi!

PR #398, opened 2019-11-06
Checked in v1beta1 go generated protos

aysylu:
This simplifies downstream integrations and release process, [see discussion](https://github.com/grafeas/grafeas/issues/380#issuecomment-544290045).


.gitignore
```diff
@@ -6,11 +6,11 @@ protoc/protoc.zip
 protoc/readme.txt
 .install.tools
 proto/**/.done
-proto/v1beta1/*_go_proto/
```
noahklein: Why only include v1beta1? What about v1?

aysylu: This is because the Grafeas server currently only uses `v1beta1` protos. Yes, we'll add `v1` go-generated protos once the server has moved to using the new protos.

PR #400, opened 2019-11-08
Effective severity should be output only.

vtsao:
It can be resolved at occurrence creation time from the note information and hence filled in during occurrence creation by the API.


vtsao:
WANT_LGTM=any

PR #402, opened 2019-11-15
Update occurrences and notes

judavi:
Hello @aysylu 
I'm submitting this fix because the parameters needs to be in order otherwise the note and the occurrence will not be updated. 
Thanks!


aysylu:
Thanks for the fix, @judavi!

PR #404, opened 2019-11-19
Add API config annotations

noahdietz:
This PR adds the API client config annotations per aip.dev, based on the [v1beta1/containeranalysis_gapic.yaml](https://github.com/googleapis/googleapis/blob/5af83f47b9656261cafcf88b0b3334521ab266b3/google/devtools/containeranalysis/v1beta1/containeranalysis_gapic.yaml). These annotations are used by client library generators to generate API client libraries. They are replacing the GAPIC config yaml file.

@vtsao this is the follow up PR to the internal CL.

### DIFF
--- a/proto/v1beta1/grafeas.proto
+++ b/proto/v1beta1/grafeas.proto
@@ -22,6 +22,9 @@ option java_package = "io.grafeas.v1beta1";
 option objc_class_prefix = "GRA";
 
 import "google/api/annotations.proto";
+import "google/api/client.proto";
+import "google/api/field_behavior.proto";
+import "google/api/resource.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/field_mask.proto";
 import "google/protobuf/timestamp.proto";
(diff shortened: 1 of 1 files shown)


proto/v1beta1/grafeas.proto
```diff
@@ -171,11 +188,17 @@ service GrafeasV1Beta1 {
...
       get: "/v1beta1/{parent=projects/*}/occurrences:vulnerabilitySummary"
     };
+    option (google.api.method_signature) = "parent,filter";
   };
 };
 
 // An instance of an analysis type that has been found on a resource.
 message Occurrence {
+  option (google.api.resource) = {
+    type: "grafeas.grafeas.io/Occurrence"
```
noahdietz: I guessed on this service name here. I'd appreciate some guidance from the team. More info can be found [here](https://aip.dev/123#terminology).

vtsao: This is fine.

noahdietz: Based on the conventions adopted by Google and Kubernetes jointly, defined [here](https://github.com/kubernetes/community/blob/c1745f728f9eda89d506134fba96adebb35307b3/contributors/devel/sig-architecture/api-conventions.md), I changed it to the top-level domain, removing the stuttering subdomain.

PR #406, opened 2019-12-02
Switch tests using cmp.Diff to use proto.Equal.

vtsao:
cmp.Diff isn't guaranteed to compare protos correctly without proto.Equal specified as an option. cmp.Ignore doesn't work with proto.Equal, so we hack it by setting the output only fields we do not want compared to their zero values.

In order to use the proper ignore we need to migrate to the Golang V2 proto bindings library (https://github.com/protocolbuffers/protobuf-go), which is not stable yet.


go/v1/api/note_test.go
```diff
@@ -38,13 +39,13 @@ func TestCreateNote(t *testing.T) {
...
-	if err := g.CreateNote(ctx, req, n); err != nil {
-		t.Errorf("Got err %v, want success", err)
+	createdNote := &gpb.Note{}
+	if err := g.CreateNote(ctx, req, createdNote); err != nil {
+		t.Fatalf("Got err %v, want success", err)
 	}
 
-	opt := cmp.FilterPath(func(p cmp.Path) bool { return p.String() == "Name" }, cmp.Ignore())
-	if diff := cmp.Diff(req.Note, n, opt); diff != "" {
+	createdNote.Name = ""
```
aysylu: Could you please add a TODO here with the note that we need to migrate to another tool? Like you explained in the PR description.

vtsao: Done.

Issue #418, opened 2020-02-25
I somehow get "Failed to paginate projects" just by curling my grafeas Projects


stubifox:
## Expected Behavior
listing of projects
## Actual Behavior
{
  "error": "Failed to paginate projects",
  "message": "Failed to paginate projects",
  "code": 13,
  "details": []
}

## Steps to Reproduce the Problem
curling the projects.
Odd is that somehow if I'm curling a specific project the error is gone and i can look up notes of the project. When i want to have the occurrences of given project, the error occurs again.

I'm running grafeas with Pgsql on Kubernetes

1.  just a simple curl to grafeas .../projects

## Additional info
I didn't do anything special. Just testing out a bit and creating notes and occurrences in a example project just using curl.
I'm pretty sure it is a DB-Issue, but don't know how to fix this without resetting the DB.

wkozlik:
A way around this error is to specify the value for `paginationkey` in the `config.yaml` file. For example:
```
paginationkey: "nhdR6G9RQ3bf5K28Rd5HFq2Qz4RJVuvbqQDLQjs75dA="
```

PR #421, opened 2020-03-09
Generating pagination key when one is not provided by the configuration.

wkozlik:
Automatic generation of pagination key had been implemented in the past, but got removed during some refactoring. https://github.com/grafeas/grafeas/blob/cd06912f00d1f23b49ecb82f39e9283625b6ec87/go/config/config.go#L79

As an alternative we could make the pagination key a required field in the configuration file: 
https://github.com/grafeas/grafeas/blob/master/go/v1beta1/config.yaml#L33 but that seems heavy-handed.

### DIFF
--- a/go/v1beta1/storage/pgsqlstore.go
+++ b/go/v1beta1/storage/pgsqlstore.go
@@ -16,6 +16,7 @@ package storage
 
 import (
 	"database/sql"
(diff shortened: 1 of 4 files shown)


aysylu:
Fixes #418

Issue #246, opened 2019-01-15
Update docs/create_note_occurrences


aysylu:
## Expected Behavior
Update the examples to use Go client library and refer to the v1beta Grafeas API reference.

## Actual Behavior
Currently, refers to Java client library from alpha version and uses [v1alpha1 Grafeas API reference](v1alpha1/docs/GrafeasApi.md)

greendinosaur:
As mentioned on other issues, I'm happy to pick this up. I can do this in two steps:

1. update the documentation with some additional JSON examples for different notes and occurrences
2. I've enhanced the [grpc example file ](https://github.com/grafeas/grafeas/blob/master/go/v1beta1/example/client/client.go)

I can submit these and in parallel, look at the Go client library and provide examples using that.

aysylu:
That'd be fantastic, @greendinosaur! Please feel free to send the PRs in an order that makes sense to you, and we'd be happy to review. Feel free to add @wkozlik as the reviewer and CC me.

PR #427, opened 2020-03-26
updating json examples and extended the grpc client example

greendinosaur:
Updates to documentation as per https://github.com/grafeas/grafeas/issues/246
I've updated the json examples as per previous PR feedback.
I've also extended the grpc client example code to create notes and occurrences. Appreciate this is a long method but seemed most appropriate to leave it like this with comments than split into smaller functions.

### DIFF
--- a/go/v1beta1/example/client/client.go
+++ b/go/v1beta1/example/client/client.go
@@ -18,18 +18,167 @@ import (
 	"context"
 	"log"
 
+	"time"
+
+	timestamp "github.com/golang/protobuf/ptypes/timestamp"
+	buildpb "github.com/grafeas/grafeas/proto/v1beta1/build_go_proto"
+	deppb "github.com/grafeas/grafeas/proto/v1beta1/deployment_go_proto"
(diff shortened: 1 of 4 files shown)


go/v1beta1/example/client/client.go
```diff
@@ -38,6 +187,176 @@ func main() {
...
+							Url:        "the git repo url",
+							RevisionId: "git commit hash",
+						},
+					},
+				},
+			},
+			TriggerId:      "triggered by code commit 123",
+			BuilderVersion: "some version of the builder",
+		},
+		ProvenanceBytes: "Z3JhZmVhcw==",
```
th3w4y: @greendinosaur not sure how to interpret buildProvenanceBytes... per documentation is the json.Marshal of buildProvenance, but is it base64 encoded or not, what is your understanding here?

perhaps @aysylu  or @wkozlik can clarify...

aysylu: Yes, [comment](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/build.proto#L87) specifies that it's base64 encoded. Does this answer your question?

aysylu (approved):

PR #432, opened 2020-04-05
Validate public key IDs use an RFC3986 conformant

steakunderscore:
The spec (protobuf) specifies that keys MUST be RFC3986 conformant. This
patch adds a validator to ensure this.


steakunderscore:
@wkozlik No problem. I noticed that container analysis API seems to be giving bad data. So figured if it's using Grafeas also, then it should probably not accept bad data.

PR #436, opened 2020-04-23
[patch] secrets were misconfigured so that certificates became unavailable

liuplgtm:
see [here](https://github.com/grafeas/grafeas/blob/master/grafeas-charts/templates/deployment.yaml#L67) and [here](https://github.com/grafeas/grafeas/blob/master/grafeas-charts/templates/deployment.yaml#L35).

the secrets configurations were conditioned on persistence.enabled=true.
This is incorrect. The server with in-memory store should also be able to use the https certificate.

To fix the problem, we simply eliminate the dependency on persistence.enabled=true.
see the commits for details pls.


wkozlik:
LGTM

PR #437, opened 2020-04-29
code refactoring to encapsulate validation into a module.

liuplgtm:
The goal is to achieve some nice consistency:
If a message A is defined by B.proto, then ValidateA should happen in validators/B.go.

for example, previously, 
ValidateCreateNoteRequest is scattered in CreateNote(), which includes:
- https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L42
- https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L45
- https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L48.
In contrast, we abstract a separate API ValidateCreateNoteRequest to encapsulate them. 
All future changes about validation will be made towards ValidateCreateNoteRequest, which is limited to the package validators/*.go.

Note: NO validation logic improvement is done, i.e., the rules shall stay equivalent to the previous ones.

Tests in note_test.go and occurrence_test.go under v1beta1/api passed. These tests cover ValidateCreateNoteRequest above.

We have not touched v1/api yet.


liuplgtm:
the only uncertainty in my mind is that:
as per the above example, 
ValidateCreateNoteRequest is scattered in CreateNote(), which includes:

https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L42
https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L45
https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L48.

Previously, 
https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L42
https://github.com/grafeas/grafeas/blob/master/go/v1beta1/api/note.go#L45
were not guarded by g.EnforceValidation flag, but after being merged to a single api call  ValidateCreateNoteRequest, they are guarded by g.EnforceValidation flag.

liuplgtm:
TODO (after the offline discussion with vtsao):
- access checking happens before validation.
- the validation of request fields like note_id should not be guarded by EnforceValidation, it should happen unconditionally. Only the validation of Note/Occurrences needs to be guarded. 
think about how to change the above code.

liuplgtm:
status update: put this **on hold**
will leave this to later, to avoid breaking stuff now.

PR #439, opened 2020-05-06
batch 1.2: validate BatchCreatenotesRequest

liuplgtm:
notes[$i].$key 
-  length ≤ 128
- format: URL-friendly check

- updated proto comment, code, test.


liuplgtm:
will merge this after batch 1.1 is done.

PR #440, opened 2020-05-07
batch 1.3: validate occurrence (generic)

liuplgtm:
remediation:
-  length ≤ 5000

resource_uri: (available in v1, but not available in v1beta) 
-  length ≤ 2048


liuplgtm:
will merge after batch 1.2

PR #443, opened 2020-05-12
Add JWT support to API

alexcope:


proto/v1/attestation.proto
```diff
@@ -69,4 +76,14 @@ message AttestationOccurrence {
   // `signature` verifies `serialized_payload`.  See `Signature` in common.proto
   // for more details on signature structure and verification.
   repeated Signature signatures = 2;
+  // One or more JWTs encoding a self-contained attestation.
+  // Each JWT encodes the payload that it verifies within the JWT itself,
+  // meaning the `serialized_payload` field is IGNORED when verifying these
```
aysylu, line 81: Somewhere in this description, let's reflect that this is the recommendation for the verifier implementations, as the OSS Grafeas server itself won't provide such guarantees, so it'll be helpful for those using this in the future to know how to correctly implement this (as opposed to assuming that the server will do this for them).

aysylu: specifically, let's say here "verifier implementation should IGNORE `serialized payload` when verifying ..., to clarify the agency on each

aysylu: the rephrasing lgtm

PR #444, opened 2020-05-15
Revert "Validate public key IDs use an RFC3986 conformant"

alexcope:
Reverts grafeas/grafeas#432

This change breaks backwards compatibility and was based on inaccurate documentation.

1. The key IDs stored by Binauthz are unfortunately not in this format. To include this change, Binauthz would either need to backfill our database or edit key IDs on the fly as we retrieve them from the database. Other users of Grafeas will likely run into similar issues.

2. Patrick Lawson originally wrote this documentation, and I've confirmed with him that "public key ID MUST be RFC3986 conformant" is an error. It should read "public key ID SHOULD be
RFC3986 conformant". I'll update the documentation after this rollback is approved.

### DIFF
--- a/go/v1/api/validators/attestation/attestation.go
+++ b/go/v1/api/validators/attestation/attestation.go
@@ -19,7 +19,6 @@ package attestation
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	gpb "github.com/grafeas/grafeas/proto/v1/grafeas_go_proto"
 )
@@ -70,8 +69,6 @@ func validateSignatures(signatures []*gpb.Signature) []error {
 	for _, s := range signatures {
 		if s.GetPublicKeyId() == "" {
 			errs = append(errs, errors.New("public key ID is required"))
-		} else if !strings.ContainsRune(s.GetPublicKeyId(), ':') {
-			errs = append(errs, errors.New("public key ID must be an RFC3986 conformant URI"))
 		}
 		if s.GetSignature() == nil {
 			errs = append(errs, errors.New("signature is required"))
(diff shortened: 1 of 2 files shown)


aysylu:
Could you please clarify the distinction made here with MUST vs SHOULD?

Also, please include the author of the original PR in this discussion.

alexcope:
MUST and SHOULD are being used as defined in [rfc2119](https://www.ietf.org/rfc/rfc2119.txt).

aysylu:
cc @steakunderscore the author of the contribution

aysylu:
> MUST and SHOULD are being used as defined in [rfc2119](https://www.ietf.org/rfc/rfc2119.txt).

got it, so it's the difference between RECOMMENDED (SHOULD) and REQUIRED

aysylu:
Approving and merging this to unblock the team, but let's continue the discussion on this thread if there's no consensus yet at this point.

steakunderscore:
@aysylu Would it not be a good idea to keep the fixes to the tests (but swapping the logic to what @alexcope expects).

Also it looks like the docs haven't been fixed[1][2]. @alexcope Since you know and have spoken to the author would it make sense for you to push a patch fixing this?

[1] :https://github.com/grafeas/grafeas/blob/master/proto/v1/common.proto#L87 
[2]: https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/common.proto#L87

aysylu:
@steakunderscore: I agree that the fixes in the tests would be great to keep. I was hasty to approve the revert of those, I apologize.

@alexcope: could you please send a follow-up PR keeping the test changes and fixing the docs?

PR #445, opened 2020-05-18
Update documentation on public key format and re-add tests

alexcope:
The `public_key_id` SHOULD be an RFC3986 conformant URI. We can not enforce any public key format without breaking backwards comparability. 

MUST and SHOULD are being used as defined in [rfc2119](https://www.ietf.org/rfc/rfc2119.txt).


aysylu:
Could we please also address the revert to the fixes to tests as suggested by @steakunderscore in https://github.com/grafeas/grafeas/pull/444#issuecomment-629505126?

proto/v1/common.proto
```diff
@@ -84,7 +84,7 @@ message Signature {
 
   // The identifier for the public key that verifies this signature.
   //   * The `public_key_id` is required.
-  //   * The `public_key_id` MUST be an RFC3986 conformant URI.
+  //   * The `public_key_id` SHOULD be an RFC3986 conformant URI.
```
aysylu, line 87: Let's also make the same update to the v1beta1/common.proto

steakunderscore:
I assume you'll need to regenerate the committed go code and swagger output:
```
proto/v1beta1/common_go_proto/common.pb.go
proto/v1beta1/swagger/grafeas.swagger.json
```

alexcope:
> I assume you'll need to regenerate the committed go code and swagger output:
> 
> ```
> proto/v1beta1/common_go_proto/common.pb.go
> proto/v1beta1/swagger/grafeas.swagger.json
> ```

done, and re-added the tests from pr #432

proto/v1beta1/vulnerability_go_proto/vulnerability.pb.go
```diff
@@ -379,8 +379,10 @@ type Vulnerability_WindowsDetail_KnowledgeBase struct {
 func (m *Vulnerability_WindowsDetail_KnowledgeBase) Reset() {
 	*m = Vulnerability_WindowsDetail_KnowledgeBase{}
 }
-func (m *Vulnerability_WindowsDetail_KnowledgeBase) String() string { return proto.CompactTextString(m) }
```
alexcope, line 382: I'm not sure why this is formatting slightly differently for me. Should I manually revert the changes to this file?

aysylu: This is strange, yes. Go formatting doesn't really restrict the number of chars per line. Let's leave this as is and come back if it starts causing issues, which I don't think it will. Thanks for flagging this!

aysylu: Also, swagger sometimes just changes the code it generates or the formatting, so you may see a similar thing in the future. It's usually fine to let it do its thing.:)

PR #447, opened 2020-05-22
Mysql Support with Filtering

birdmanja:
Added mysql support and filtering for mysql.


aysylu:
Hi @birdmanja,

Thanks for your contribution, looking forward to the review. Could you please ping this thread once the other issues are fixed? Then @wkozlik and @liuplgtm will be able to review this. Also, do you think it'd be possible to split this into a separate DB integration? This is where the project is moving long term. See examples for [Oracle DB](https://github.com/judavi/grafeas-oracle) by @judavi and https://github.com/grafeas/grafeas-pgsql (which will be expanded for further separation).

PR #453, opened 2020-09-09
pgsqlstore: Minor code clean up

### LINKED ISSUE #452: Postgres storage pagination key error message should reference a 32 byte rather than 32 bit key
Fernet takes and generates 32 byte keys.

Thus the error message here
https://github.com/grafeas/grafeas/blob/07ec5974c3df8413c5f6f1174d860eec064d8980/go/v1beta1/storage/pgsqlstore.go#L58 is confusing and made it difficult to figure out why the pagination key I provided in config was wrong.

Updating this error message would make debugging config issues much easier.

c4milo:
Also fixes #452.


wkozlik (approved):
LGTM
Thank you for this PR.

Issue #455, opened 2020-09-17
Trying to delete a note returns marshaling error in the server


c4milo:
## Expected Behavior
gRPC client call returns without error. 

## Actual Behavior
I get the following error printing out in Grafeas logs: 

```
ERROR: 2020/09/17 19:08:53 grpc: server failed to encode response:  rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil
```

and an internal error sent back to the client. 

## Steps to Reproduce the Problem

Environment, commands

1.  Create a note
2.  Try to delete the note  

## Additional info
I used my own generated gRPC client java library. Creating notes with it works fine.

I started the Grafeas server using memstore, as follows:
```
$ go run main/main.go --config config.yaml
2020/09/17 19:08:33 starting grpc server on 0.0.0.0:8090
2020/09/17 19:08:33 grpc server is configured without client certificate authentication
```

PR #456, opened 2020-09-18
api/v1beta1: Fix marshalling error on delete calls

c4milo:
which fixes https://github.com/grafeas/grafeas/issues/455


wkozlik (approved):
LGTM. 
Thank you for this fix.

PR #454, opened 2020-09-11
grafeas.proto: Document key purpose in notes map

c4milo:

Issue #343, opened 2019-07-17
Unable to go to next page


bansalrajat:
I am unable to go to next page using the pageSize and nextPageToken .
Also the nextPagetoken is always  changing when ever i am refreshing the url . 
I am using postgres as db with grafeas 

#### Some Encoded String
"nextPageToken":"gAAAAABdLmm6ZtyRjjn5KDAWdbAHTZ40u81TbharedFdKNHCwPus26R2I9jG2EjZFp_sMA0uRGjemkx3sRTEoIah0SpQNs-1hg=="}

vtsao:
What list methods are calling?

Can you post the code you're using to do the list call?

Also, can you clarify what's happening? Are you getting an error, or not getting the results you expect based on what's in your database?

aysylu:
Closing this due to inactivity. Please re-open if this is still an issue.

PR #458, opened 2020-10-20
add missing call to validatePageSize in ListNoteOccurrences

th3w4y:
Issue: 

the func (g *API) ListNoteOccurrences does not use validatePageSize(req.PageSize) but req.PageSize in its call to g.Storage.ListNoteOccurrences() 

This potentially results in an [] empty response due to the use of LIMIT = pageSize in the queries.go, which gets set to 0 (unspecified default req.PageSize) 

cc: @aysylu @wkozlik 

* add pageSize validation to ListNoteOccurrences
* regenerated files after comment change in PR #454
* allow cel/syntax.pb.go so that stg. engines be able to import github.com/grafeas/grafes/cel for filter implementation


th3w4y:
After the proposed fixed one can successfully use ListNoteOccurrences.

```
curl http://localhost:8080/v1beta1/projects/secret-project/notes/demo/occurrences | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11861    0 11861    0     0  2316k      0 --:--:-- --:--:-- --:--:-- 2316k
{
  "occurrences": [
    {
      "name": "projects/secret-project/occurrences/80556e60-1922-4b7b-a3d5-de674a00157b",
      "resource": {
        "name": "",
        "uri": "projects/secret-project/notes/demo",
        "contentHash": null
      },
      "noteName": "projects/secret-project/notes/demo",
      "kind": "BUILD",
      "remediation": "",
      "createTime": "2020-10-20T17:30:28.335301Z",
      "updateTime": null,
      "build": {
        "provenance": {
          "id": "demo",
          "projectId": "secret-project",
          "commands": [],
...
```
[potentially fixing the pageSize part of the behaviour described in #343]

wkozlik:
Thank you for the fix. Could you add a unit test to validate page size?

th3w4y:
this PR would only introduce consistency as all other methods defined by API are calling the validatePageSize(req.PageSize) without such unit tests...

if i am to add tests for validatePageSize the blast of the change gets way bigger, testing the empty zero value of req.pageSize would most likely require some modification of behaviour in the fakeStorage to return [] in such event, and this might affect more use-cases.


i suggest if possible merge the PR as such as it brings the consistency (validatePageSize(req.PageSize) was missed from that method in the past) then i will think of a way to verify the unit-test arround validations of zero-values generated by the protos vs the conventions value of what we would expect the defaults to be as per the comments in the .proto files.

th3w4y:
a follow up for the unit-tests https://github.com/grafeas/grafeas/issues/459

wkozlik:
The existing test for ListOccurrences (https://github.com/grafeas/grafeas/blob/f11fa45f2a57406a0b69eb3463ecbca9e127ccd2/go/v1beta1/api/occurrence_test.go#L191) covers page size validation. I was thinking about adding a similar test case in https://github.com/grafeas/grafeas/blob/f11fa45f2a57406a0b69eb3463ecbca9e127ccd2/go/v1beta1/api/occurrence_test.go#L749. Would that work?

th3w4y:
looking into the unit-test, but not sure that can be captured under TestListNoteOccurrencesErrors, as pagesize = 0 should be rewritten to defaultPageSize = 20 and will not produce an err.

wkozlik (approved):
LGTM
Thank you for the fix and being very thorough.

PR #460, opened 2020-11-09
Use subtests in occurrence_test

ezimanyi:
Updating each test case to run as a subtest makes the output clearer. This is most apparent in test cases that log errors; when all cases run as a single test, the output contains the log statements from all cases.

I came across this while making some changes that required updating these tests; submitting as a separate PR as it's completely separate.


ezimanyi:
(Strongly recommend reading ignoring whitespace; this is mostly indentation changes.)

aysylu (approved):

PR #461, opened 2020-11-10
Make occurrence update tests less dependent on implementation details

ezimanyi:
The TestUpdateOccurrenceErrors currently rely on the fact that we check for an existing occurrence last.

For example, if writing to the the storage layer is erroring *and* there is no existing occurrence, the test is asserting that we return the storage error, which seems overspecified.

Update the tests so that there is an existing occurrence in the database for the test cases that are exercising other error paths.

The one exception is that it is important to return a permission error instead of a NotFound error if the user doesn't have the required permission (to avoid leaking the existence of a resource they don't have permission to see). I've added a test case to specifically validate that case. (The code is correct here, the test is just ensuring we are specifically testing that case.)

### DIFF
--- a/go/v1/api/occurrence_test.go
+++ b/go/v1/api/occurrence_test.go
@@ -504,7 +504,8 @@ func TestUpdateOccurrenceErrors(t *testing.T) {
 
 	tests := []struct {
 		desc                        string
-		occName                     string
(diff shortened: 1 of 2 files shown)


wkozlik (approved):
Thank you for improving Grafeas!

LGTM

PR #462, opened 2020-11-15
upgrade to grpc-gateway/v2

th3w4y:
* upgrade protoc to 3.13
* upgrade grpc-gateway to v2.0.1
* use protocomp.Transform() for cmd.Diff with proto
* &runtime.JSONPb{} in v2/runtime defaults to the previously called in v1 EmitDefaults 
* use of protocmp.IgnoreFields(proto.MessageV2(&pb.Note{}), "create_time", "update_time"), instead of cmd.Path approach
* adopted v1 generate.go to work with higher tools
* adopted v1beta1 generate.go to work with higher tools
* adopted v1alpha to work with higher version tools, and to use openapi2 annotations instead of swagger (dues to grpc-gateway/v2)
* changed go/v1/{note,occurrences} and its tests to return the proper (pb.X, error) due to *resp = *n resulting in ssignment copies lock value to req (as google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex)
* updated circleci config to use Golang:1.14
* propagate context to the sql operation in the Postgresql store in preparation for driver tracing/instrumentation
* execute get count  before query/exec (in Postgresql store)  to min. open conn.

FYI: @aysylu @wkozlik


wkozlik:
Wow! That is quite a bit of work you did. I will look it over later this week.

th3w4y:
@wkozlik it only looks big, because of the autogenerated files 
the generate.go for the  protoc/swagger have given a small headache though (especially for the v1)... but the one committed now (without relative paths) is functional.

th3w4y:
quite busy times towards the end of year... 
@wkozlik got any chance so far to zoom in to my changes?

wkozlik:
Busy as well. I will try to find time to review it today or tomorrow.

wkozlik (approved):
LGTM.
Thank you for your contribution! And I apologize for the delay reviewing it.

PR #465, opened 2020-11-18
Resolution to #237 - postgres for testing

tomcruise81:


go/v1beta1/storage/pgsqlstore_test.go
```diff
@@ -15,16 +15,375 @@
...
+		if _, err := os.Stat(postgresZipFileName); !os.IsNotExist(err) {
+			postgresZipFound = true
+		}
+
+		if !postgresZipFound {
+			//URLs from https://www.enterprisedb.com/download-postgresql-binaries
+			postgresUrl := ""
+			switch runtime.GOOS {
+			case "darwin":
+				postgresUrl = "https://sbp.enterprisedb.com/getfile.jsp?fileid=1257418&_ga=2.59164973.1284458936.1605634932-924989421.1605634932"
```
th3w4y, line 286: Any reason why this files gets downloaded from EnterpriseDB
I am not sure that would be such a good idea

th3w4y:
This would silently force the end user into accepting EnterpriseDB terms and conditions https://www.enterprisedb.com/terms-use-gated which I don't think is a nice thing to do.

th3w4y:
Also the link url that you provide seems to use analytics linked to one particular customer registration on EnterpriseDB side, which would actually mean that privacy information about me as an end user would be collecting about EnterpriseDB and the only person that could seek that information deletion under the GDPR of that information would be the person that owns that (ga=FANCY_TRACKER) which might mean you in that case.

Also the owner of that ga can request its data from EnterpriseDB and then collected analytics about people that execute the Unit tests, which I don't think it would be something appropriate.

Thanks.

@wkozlik  @aysylu  FYI, I think is not safe to approve this PR as such.

go/v1beta1/storage/pgsqlstore_test.go
tomcruise81: Nothing nefarious was meant. Just got the download links from https://www.postgresql.org/download/windows/ → [zip archive](https://www.postgresql.org/download/windows/#pgContentWrap:~:text=Advanced%20users%20can%20also%20download%20a%20zip%20archive%20of%20the%20binaries%2C%20without%20the%20installer). The intent was to have a locally accessible binary without requiring developers to do a full PostgreSQL installation.

If there are other links that you'd recommend, I'm totally open to suggestions.

tomcruise81:
> This would silently force the end user into accepting EnterpriseDB terms and conditions https://www.enterprisedb.com/terms-use-gated which I don't think is a nice thing to do.

Not sure why it would force end-user's into accepting it. It might force developers who are actually downloading it when they run the tests into accepting it, which I agree isn't really desirable. I'm happy for suggestions as to alternative download links.

tomcruise81:
> Also the link url that you provide seems to use analytics

@th3w4y - Would you prefer the final links (after following)? Something like:

- https://get.enterprisedb.com/postgresql/postgresql-10.15-1-windows-binaries.zip
- https://get.enterprisedb.com/postgresql/postgresql-10.15-1-windows-x64-binaries.zip
- https://get.enterprisedb.com/postgresql/postgresql-10.15-1-linux-x64-binaries.tar.gz
- etc.

th3w4y:
> > Also the link url that you provide seems to use analytics
> 
> @th3w4y - Would you prefer the final links (after following)? Something like:
> 
> * https://get.enterprisedb.com/postgresql/postgresql-10.15-1-windows-binaries.zip
> * https://get.enterprisedb.com/postgresql/postgresql-10.15-1-windows-x64-binaries.zip
> * https://get.enterprisedb.com/postgresql/postgresql-10.15-1-linux-x64-binaries.tar.gz
> * etc.

that is a different situation

go/v1beta1/storage/pgsqlstore_test.go
th3w4y: Thanks for the reply I think the direct links are way better

wkozlik:
I would think that using docker-compose suggestion from https://github.com/grafeas/grafeas/issues/237 would be more flexible. Is there a reason to not to go that path?

tomcruise81:
> I would think that using docker-compose suggestion from #237 would be more flexible. Is there a reason to not to go that path?

@wkozlik - that is definitely one way of doing it. The approach presented in this PR doesn't require `docker` to run the tests. For my purposes this worked out to be a better solution.

tomcruise81:
To elaborate, the initial issue stated:

> The test should pass locally by bringing up a local Postgres instance to connect to.

The goal of this PR was to do that, without introducing the additional requirement of `docker` just to run the tests locally. However, I'm also willing to strip out all of the downloading code, and simplify this PR to just do two things:

1. Check for a running instance of Postgres that the tests have access to (i.e. have credentials for and can connect to)
2. If no instance is running, _but_ the system contains applicable Postgres binaries (i.e.`initdb` and  `pg_ctl`) accessible via the system PATH, then to initialize, startup & teardown a testing Postgres instance as part of testing.

Then if someone else via a separate PR wants to implement the docker-compose suggestion from #237, item 1 above would allow the tests to keep working. Item 2 would be for folks that just want to run `go test ./...` without having to bring `docker` into the mix

wkozlik:
I like your suggestion about stripping out the downloading code. I can look into creating a docker-compose file separately.

tomcruise81:
@wkozlik - done.

wkozlik (approved):
LGTM

wkozlik:
Thank you for the big improvement!

PR #466, opened 2020-12-03
Improves description of NOTE_KIND_UNSPECIFIED

geriom:
Provide a better description `NOTE_KIND_UNSPECIFIED`,  that is never actually used, but `NoteKind` always a required field.


wkozlik (approved):
LGTM

th3w4y:
shouldn't this be actually updated here? https://github.com/grafeas/grafeas/blob/ff50c7e607bcf4c9e4586d667b38c1f2fb067c05/proto/v1beta1/common.proto#L26

wkozlik:
good catch. I'll update it

PR #472, opened 2021-01-08
Add source and inclusive fields to v1 and v1beta1 proto.

nms2140:
Adding the source field to the Detail and inclusive boolean to the Version proto. This PR only adds it to v1.


proto/v1/vulnerability.proto
```diff
@@ -114,6 +114,9 @@ message VulnerabilityNote {
     // upstream timestamp from the underlying information source - e.g. Ubuntu
     // security tracker.
     google.protobuf.Timestamp source_update_time = 12;
+
+    // The source from which the information in this Detail was obtianed.
```
wkozlik, line 118: nit: typo in "obtianed"

wkozlik (approved):
LGTM

PR #473, opened 2021-01-21
Update to curl flags with certificate

dinagraves:
The instructions for using the curl command with the certificate didn't work for me.  I got the error message: 

`curl: (58) unable to set private key file: 'server.pem' type PEM`

When I added these additional flags, it worked fine.


wkozlik (approved):
Thank you.

PR #476, opened 2021-03-22
Add the vendor field to Vulnerability.Detail

vyinpengli:
Add the vendor field to Vulnerability.Detail, which is necessary to differentiate between products of the same name in the NVD 1.1 feed.


vyinpengli:
@googlebot I signed it!

wkozlik (approved):
LGTM

PR #478, opened 2021-04-28
fixes ARM build

catherinejones:
Building an ARM version of Grafeas fails presently, because the protoc version installed is for x86_64. This updates the protoc download script to pull down the arm64 version if appropriate!


wkozlik (approved):

wkozlik:
Thank you for this fix!

PR #479, opened 2021-04-30
initial commit of compliance Occurrence and Note

nms2140:
Adds a compliance occurrence and Note.


proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 The Grafeas Authors. All rights reserved.
```
aysylu, line 1: 21

proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
...
+import "proto/v1/package.proto";
+import "proto/v1/vulnerability.proto";
+
+option go_package = "github.com/grafeas/grafeas/proto/v1/grafeas_go_proto";
+option java_multiple_files = true;
+option java_package = "io.grafeas.v1";
+option objc_class_prefix = "GRA";
+
+message ComplianceNote {
+  // The OS and config versions the benchmark applies to.
```
aysylu, line 30: is 1 intentionally skipped?

proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
...
+  repeated grafeas.v1.ComplianceVersion version = 2;
+  // The title that identifies this compliance check.
+  string title = 3;
```
aysylu, line 33: let's put title and description as the two first fields, then the version (per convention)

proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
...
+  // The CPE URI (https://cpe.mitre.org/specification/) this benchmark is
+  // applicable to.
+  string cpe_uri = 1;
+  // The version of the benchmark. This is set to the version of the OS-specific
+  // CIS document the benchmark is defined in.
+  string version = 2;
+}
+
+message ComplianceOccurrence {
+  repeated NonCompliantFile non_compliant_files = 2;
```
aysylu, line 64: could we add a short description clarifying that this occurrence is to be generated only when the requirements in ComplianceNote aren't satisfied for an artifact?

proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+message NonCompliantFile {
+  string path = 1; // Empty if `display_command` is set.
```
aysylu, line 67: what does `display_command` mean in the context of files? i.e. why won't the path be set if the `display_command` is set? this seems like something an implementation could decide to enforce or not

aysylu (changes requested):

proto/v1/compliance.proto
erikvarga: There should be an "id" field before this with the number 1 as per the design.

proto/v1/compliance.proto
```diff
@@ -0,0 +1,80 @@
...
+// An indication of how severe a violation of a compliance check is.
+enum Severity {
+  UNSPECIFIED = 0;
+  LOW = 1;
+  MEDIUM = 2;
+  HIGH = 3;
+  CRITICAL = 4;
+}
+
+message ComplianceOccurrence {
```
erikvarga, line 63: There be a "string id = 1;" here too as per the design.

proto/v1/compliance.proto
erikvarga: display_command is the command that can be used to display the non-compliant files. For example, for the benchmark "6.1.10 Ensure no world writable files exist", we could list all files on the filesystem that were found to be world-writable:
```
[{path: "/home/bob/file1.txt", reason: "File's +w bit set for all users."},
 {path: "/home/bob/file2.txt", reason: "File's +w bit set for all users."},
 {path: "/home/bob/file3.txt", reason: "File's +w bit set for all users."},
 ...]

```
But we could also just show this one command:
```
{display_command: "df --local -P | awk '{if (NR!=1) print $6}'
        | xargs -I '{}' find '{}' -xdev -type f -perm -0002",
 reason: "One or more files have the +w bit set for all users."}
```

When this command is set, it doesn't make much sense to list the paths individually since the command already defines the files.

erikvarga (changes requested):

proto/v1/compliance.proto
nms2140: Done.

proto/v1/compliance.proto
nms2140: I did not include the id since our general pattern is to store the occurrenceId / note Id in the grafeas proto. https://github.com/grafeas/grafeas/blob/master/proto/v1/grafeas.proto#L194

Lmk if you think we still should include it here. Similar reason for the Note as well.

proto/v1/compliance.proto
nms2140: I omitted it because the Occurrence proto has a way to set id: https://github.com/grafeas/grafeas/blob/master/proto/v1/grafeas.proto#L194

proto/v1/compliance.proto
nms2140: Done.

proto/v1/compliance.proto
nms2140: Done.

proto/v1/compliance.proto
```diff
@@ -0,0 +1,71 @@
...
+  string version = 2;
+}
+
+message ComplianceOccurrence {
+  repeated NonCompliantFile non_compliant_files = 2;
+}
+message NonCompliantFile {
+  string path = 1; // Empty if `display_command` is set.
+  string display_command = 2; // Command to display the non-compliant files.
+  string reason = 3;
```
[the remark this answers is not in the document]
nms2140: Done.

proto/v1/compliance.proto
erikvarga: We don't store the whole Note/Occurrence in the CIS scanner, just the ComplianceNote/Occurrence part of it, and for the instance scanning integration, we, we also just look at the ComplianceNotes/Occurrences. So there we'd need the ID in this message in order to connect the non-compliance report with the benchmark that was found non-compliant.

proto/v1/compliance.proto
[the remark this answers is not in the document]
aysylu: sorry I don't see it, perhaps it wasn't pushed to repo?

proto/v1/compliance.proto
aysylu: Got it, thank you! Would it be possible to capture some of this explanation in a brief comment, for clarity?

aysylu (changes requested):

proto/v1/compliance.proto
nms2140: We decided to create a new message / struct to return the compliance occurrence and id.

proto/v1/compliance.proto
nms2140: Had a discussion offline, and we are agreeing to use a different struct / message to store the Id.

proto/v1/compliance.proto
nms2140: Done.

aysylu (approved):

PR #490, opened 2021-07-22
Initial commit of DSSE Attestation and IntotoProvenance support.

sherzberg-1:


proto/v1/common.proto
```diff
@@ -103,3 +105,17 @@ message Signature {
   //   * "nih:///sha-256;703f68f42aba2c6de30f488a5ea122fef76324679c9bf89791ba95a1271589a5"
   string public_key_id = 2;
 }
+
+// MUST match https://github.com/secure-systems-lab/dsse/blob/master/envelope.proto.
+// An authenticated message of arbitrary type.
+message Envelope {
+  message Signature {
```
sherzberg-1, line 112: Note that common.proto already has a Signature field. So I moved the Envelope Signature inside the Envelope message, which is technically different than https://github.com/secure-systems-lab/dsse/blob/master/envelope.proto.

wkozlik (approved):
LGTM

proto/v1/common.proto
aysylu: Yes, this sounds great. In v2, we'll make sure to reconcile these.

proto/v1/dsse_attestation.proto
```diff
@@ -0,0 +1,66 @@
...
+    string human_readable_name = 1;
+  }
+  // Hint hints at the purpose of the attestation authority.
+  Hint hint = 1;
+}
+
+message DsseAttestationOccurrence {
+  // If doing something security critical, make sure to verify the signatures in this metadata.
+  Envelope envelope = 1;
+  oneof decodedPayload {
```
aysylu, line 46: nit: snake case, `decoded_payload`

proto/v1/grafeas.proto
```diff
@@ -239,6 +240,10 @@ message Occurrence {
     grafeas.v1.UpgradeOccurrence upgrade = 15;
     // Describes a compliance violation on a linked resource.
     grafeas.v1.ComplianceOccurrence compliance = 16;
+    // https://github.com/secure-systems-lab/dsse
+    grafeas.v1.DsseAttestationOccurrence dsse_attestation = 17;
+    // https://github.com/secure-systems-lab/dsse
```
aysylu, line 245: `envelope` should be in a top-level of the Occurrence, not as a `oneof` field

proto/v1/intoto_provenance.proto
```diff
@@ -0,0 +1,87 @@
...
+
+package grafeas.v1;
+
+option go_package = "github.com/grafeas/grafeas/proto/v1/grafeas_go_proto";
+option java_multiple_files = true;
+option java_package = "io.grafeas.v1";
+option objc_class_prefix = "GRA";
+
+import "google/protobuf/timestamp.proto";
+
```
aysylu, line 25: could you pls add the link to the spec? https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md

proto/v1/dsse_attestation.proto
```diff
@@ -0,0 +1,66 @@
...
+    InTotoStatement statement = 2;
+  }
+}
+// The serialized InTotoStatement will be stored as Envelope.payload.
```
aysylu, line 50: could you pls add a link to the spec https://github.com/in-toto/attestation/blob/main/spec/README.md#statement?

aysylu (changes requested):

proto/v1/dsse_attestation.proto
sherzberg-1: Done

proto/v1/intoto_provenance.proto
sherzberg-1: Done

proto/v1/grafeas.proto
sherzberg-1: Ah, good catch, thanks.

aysylu (approved):

PR #494, opened 2021-07-27
Renamed Builder to BuilderConfig

sherzberg-1:
Renamed Builder to BuilderConfig, to not conflict with the Java generated Builder class.

### DIFF
--- a/proto/v1/intoto_provenance.proto
+++ b/proto/v1/intoto_provenance.proto
@@ -71,12 +71,12 @@ message Metadata {
   bool reproducible = 5;
 }
 
-message Builder {
+message BuilderConfig {
   string id = 1;
 }
 
 message InTotoProvenance {
-  Builder builder = 1; // required
+  BuilderConfig builder_config = 1; // required
   // Identifies the configuration used for the build.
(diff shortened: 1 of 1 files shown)


proto/v1/intoto_provenance.proto
```diff
@@ -71,12 +71,12 @@ message Metadata {
   bool reproducible = 5;
 }
 
-message Builder {
+message BuilderConfig {
```
aysylu, line 74: nit: maybe BuilderIdentity? since the submessage has the ID only, vs configuration of the builder. open to other ideas as well

aysylu (changes requested):

proto/v1/intoto_provenance.proto
sherzberg-1: Named it that based on the comment:
// Identifies the configuration used for the build.
But happy to rename it. If it gets a new field, it may be more than Identity.

aysylu: sgtm, makes sense

aysylu (approved):

PR #498, opened 2021-08-09
Move the package type and effective severity to the PackageIssue level

nyc:
When we are creating vulnerabilities, we may find vulnerabilities for
multiple different package types (e.g. OS, Maven, NPM, Go, etc.) on a
single Note. Since type is currently on the Vulnerability occurrence
level, we need to move it down into the PackageIssue level. Similarly,
vulnerabilities issued for different package types (potentially by
different vulnerability sources) can have different effective
severities. Therefore, this field also needs to be moved down into the
PackageIssue level.


aysylu (approved):

aysylu:
@wkozlik this LGTM, could you please also take a look before we merge? thank you!

wkozlik (approved):
LGTM

PR #504, opened 2021-09-28
Added field annotations for output-only fields.

wkozlik:


proto/v1beta1/spdx.proto
```diff
@@ -377,33 +407,3 @@ enum RelationshipType {
...
-// https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/
-message RelationshipNote {
-  // The type of relationship between the source and target SPDX elements
-  RelationshipType type = 1;
-}
-
-// RelationshipOccurrence represents an SPDX Relationship
-// section:
-// https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/
-message RelationshipOccurrence {
```
aysylu, line 392: is this intentionally deleted?

aysylu (changes requested):

proto/v1beta1/spdx.proto
wkozlik: This message is still there. I just moved enum to the bottom of the file.

aysylu: Oh I see now, apologies, I missed it in the diff earlier.

aysylu (approved):

PR #510, opened 2021-10-28
add support for keyType in signature by adding a new message type

hkadakia:
This pull request allows users to specify the key type used in the public key id field for the signature. By mentioning the type it provides more structure to the Signature message & also adds attributes to the signature.


hkadakia:
@aysylu Should I update the PR to add the KeyType field to the existing Signature proto.

aysylu:
@hkadakia let's get input from @patricklawsongoogle on this, since his team is using this `Signature` as well, so want to make sure this won't negatively impact anything there.

patricklawsongoogle:
First: I'm not sure how relevant v1beta1 API surface is, so I'll let Aysylu speak to that.  My comments below apply primarily to Grafeas v1 and future revisions (e.g. v2).

I'm going to push back on this for a few reasons:

* First and foremost, we're trying to standardize all signatures to [DSSE](https://github.com/secure-systems-lab/dsse).  DSSE is flexible enough to sign effectively anything with minimal dependencies, and includes several important security hardening features at the design level.  It's also a central place to make suggestions and improvements to the signature protocol and ensure that security engineers from multiple organizations have reviewed the changes before they go into effect.  For this reason, I don't think we should add any more signature envelopes to Grafeas; instead, we should encourage users to switch to DSSE, and if it's missing a feature they need, to propose that feature on DSSE itself.
* On adding the signing algorithm to the signature message: we deliberately did not do this in DSSE--we consider it to be a security usability issue.  Transmitting the signing algorithm in-band but unauthenticated implies to verifiers that they should trust that field, but they shouldn't.  An attacker could change the transmitted signing algorithm, and the verifier wouldn't know unless they had the signing algorithm pre-configured--and they *should* have the algorithm preconfigured from a trusted source along with the public key, both of which are identified by the key ID.  And since they should have it preconfigured anyway, there's no need for the in-band signing algorithm; it can only do harm by misleading a verifier into thinking that it should be trusted.  For an extreme example of this type of issue being exploited in the wild, see the [JWT none algorithm vulnerability](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/).

hkadakia:
Thanks for a detailed explanation Patrick. If Grafeas community recommends to start using `DSSE` for signing artifacts, we can incorporate those changes in our local repo.
Aysylu can you please let me know if the proposed change above can go in v1Beta1 or I can close my PR as not needed.
Thanks for taking a look.

aysylu:
@hkadakia: To stay consistent with the best practices, I recommend closing this PR. However, if you'd like to adopt the use of [DSSE Envelope for signed metadata](https://github.com/grafeas/grafeas/blob/master/proto/v1/grafeas.proto#L248) and use it v1beta1, happy to review the PR that brings it in this version.

PR #512, opened 2021-11-01
Add in the proto changes required for adding cvss v3 info to Occs.

nms2140:


proto/v1/vulnerability.proto
```diff
@@ -174,6 +174,17 @@ message VulnerabilityOccurrence {
   // severity.
   float cvss_score = 3;
 
+  // The CVSS v3 score for this vulnerability.
+  message CVSSV3 {
+    // The base score for for this vulnerability according to cvss v3.
+    float base_score = 1;
+    // The severity rating assigned to this vulnerability by vulnerability provider.
+    Severity severity = 2;
```
aysylu, line 182: I thought we wanted to expose only base_score for now? Did the understanding change?

aysylu (changes requested):

proto/v1/vulnerability.proto
nms2140: yes there was a thread with Shmuel and Brian and we agreed to add in the severity as well.

aysylu: Commented on the doc as I don't see the mention of it anywhere. SGTM with the overall approach to record it.

aysylu (approved):

PR #513, opened 2021-11-01
introduce a common CVSS proto message which can be supported for both v2 & v3

hkadakia:
Use-case: When extracting information from jfrog xray payload, it contains info for CVSSv3, CVSSv2, CVE & CWE. The current `cvss.proto` & `vulnerability.proto` file does not support CVSSv2 format. In order to add xray payload to grafeas vulnerability notes, we need to introduce the new field in vulnerability message.


proto/v1beta1/vulnerability.proto
```diff
@@ -146,7 +146,14 @@ message Vulnerability {
   // security tracker.
   google.protobuf.Timestamp source_update_time = 6;
 
-  // Next free ID is 7.
+  // This field tracks the CVE record for this vulnerability
+  string cve = 7;
```
aysylu, line 150: We already record this at the top-level, when we attach the Occurrence to the note and record the CVE number in `note_name`: https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/grafeas.proto#L214. Does this help?

proto/v1beta1/vulnerability.proto
```diff
@@ -146,7 +146,14 @@ message Vulnerability {
...
+  // The CWE includes both software & hardware weakness types for this vulnerability.
+  repeated string cwe = 8;
```
aysylu, line 152: Are you expecting the CWEs to be treated same as CVEs? Would it make sense to attach vulnerability Occurrences to CWE notes, same as what's done for CVEs?

proto/v1beta1/vulnerability.proto
```diff
@@ -146,7 +146,14 @@ message Vulnerability {
...
+  // The full description of the CVSSv2.
+  CVSSv2 cvss_v2 = 9;
```
aysylu, line 154: We're adding CVSSv3 in #512, FYI. @nms2140 WDYT of this change to our OSS proto?

aysylu (changes requested):

proto/v1beta1/vulnerability.proto
hkadakia: Yes I think having multiple notes for each CWE & then a corr. vulnerability occurence might work. But would that mean we would have multiple vuln occurences with the same `CVSSv3` & maybe `CVSSv2`.

One other concern, if the same resource url has CVE & multiple CWEs do you see any problem when creating occurence or that should be acceptable.

proto/v1beta1/vulnerability.proto
hkadakia: The only concern as expressed below is having the same resource url across multiple occurences with different note name. (each for CVE & CWEs)

proto/v1beta1/vulnerability.proto
hkadakia: @nms2140 Any thoughts on this?

nms2140: Let me clarify this a bit:

Lets say we have a proto named CVSS:

message CVSS {
  // The base score is a function of the base metric scores.
  float base_score = 1;
  float exploitability_score = 2;
  float impact_score = 3;

  // Base Metrics
  // Represents the intrinsic and fundamental characteristics of a vulnerability that are
  // constant over time and user environments
  AttackVector attack_vector = 4;
  AttackComplexity attack_complexity = 5;
  Authentication authentication = 6;
  Impact confidentiality_impact = 7;
  Impact integrity_impact = 8;
  Impact availability_impact = 9;
  Scope scope = 9;
  PrivilegesRequired privileges_required = 7;
  UserInteraction user_interaction = 8;

  enum AttackVector {
    ATTACK_VECTOR_UNSPECIFIED = 0;
    ATTACK_VECTOR_LOCAL = 1;
    ATTACK_VECTOR_ADJACENT = 2;
    ATTACK_VECTOR_NETWORK = 3;

  }

  enum AttackComplexity {
    ATTACK_COMPLEXITY_UNSPECIFIED = 0;
    ATTACK_COMPLEXITY_HIGH = 1;
    ATTACK_COMPLEXITY_MEDIUM = 2;
    ATTACK_COMPLEXITY_LOW = 3;
    ATTACK_VECTOR_PHYSICAL = 4;
  }

  enum Authentication {
    AUTHENTICATION_UNSPECIFIED = 0;
    AUTHENTICATION_MULTIPLE = 1;
    AUTHENTICATION_SINGLE = 2;
    AUTHENTICATION_NONE = 3;
  }

  enum PrivilegesRequired {
    PRIVILEGES_REQUIRED_UNSPECIFIED = 0;
    PRIVILEGES_REQUIRED_NONE = 1;
    PRIVILEGES_REQUIRED_LOW = 2;
    PRIVILEGES_REQUIRED_HIGH = 3;
  }

   enum UserInteraction {
    USER_INTERACTION_UNSPECIFIED = 0;
    USER_INTERACTION_NONE = 1;
    USER_INTERACTION_REQUIRED = 2;
  }

  enum Scope {
    SCOPE_UNSPECIFIED = 0;
    SCOPE_UNCHANGED = 1;
    SCOPE_CHANGED = 2;
  }

  enum Impact {
    IMPACT_UNSPECIFIED = 0;
    IMPACT_HIGH = 1;
    IMPACT_LOW = 2;
    IMPACT_NONE = 3;
    IMPACT_PARTIAL = 4; 
    IMPACT_COMPLETE = 5;
  }
}

We could then represent either CVSS v2 or CVSS v3 scores with the same proto:

CVSS cvssV2 = 9;
CVSS cvssV3 = 10;

The hope is that any additional versions that are added to cvss (example: cvss v3.1), which will add yet another score will not need the addition of a new proto for representing it, but just another field, say:

CVSS cvssV3_1 = 11;

In fact, this is purely additive to the existing CVSSV3 proto, so perhaps it might make sense to add these extra fields directly in there rather than creating a new Proto. The only downside is that it is named "CVSSV3", which is going to be a little misleading.

Thoughts?

hkadakia: Creating a generic message which contains both v2 & v3 makes lot more sense & I do like the above approach since it is kinda future proof.
Adding new fields (related to v2) to "CVSSv3" could land up creating more confusion and would definitely be misleading.
If you & Aysylu agree with the above approach, I can update this PR.

nms2140: We do agree (I discussed this with her before posting it here), so please update the PR

proto/v1beta1/vulnerability.proto
aysylu: This would be WAI: we create multiple `VulnerabilityOccurrence`s for a given image, each of which is attached to a different `VulnerabilityNote` that represents a CVE. Does this help clarify the model?

proto/v1beta1/vulnerability.proto
aysylu: IIUC the model you're describing, you'd want `VulnerabilityNote` per each CVE and CWE. Then for a given affected image, you'd create `VulnerabilityOccurrence` with the same `resource_uri` and attached to the relevant `VulnerabilityNote`s that correspond to the CVEs/CWEs. The score is per `VulnerabilityOccurrence`, so yes, they may have similar CVSS v2 and v3 scores across multiple.

proto/v1beta1/vulnerability.proto
aysylu: @hkadakia: I agree with @nms2140. Re backwards compatibility: for v1beta1, there are no backwards compatibility guarantees, and this is the version used by the OSS Grafeas server implementation. The story for v1 is different, since it's required to have backwards-compatible changes only.

aysylu (changes requested):
Please see the discussion of cvss v2 vs v3 for the requested changes. Thanks!

proto/v1beta1/vulnerability.proto
nms2140: @hkadakia I am blocked on using this CVSS proto - I had been about to start off on some work involving CVSS when this PR came in (https://github.com/grafeas/grafeas/pull/512). Since I found out about this I thought I would use this new proto instead. I plan to start working on a PR to do this tomorrow so I can unblock myself.

hkadakia: I was about to start working on this PR. If you don't mind waiting for a day or two I can get that done. Let me know what works for you.

hkadakia: @nms2140 I have updated the PR based on your comments above. Let me know if there is any additional change.

proto/v1beta1/cvss.proto
```diff
@@ -21,8 +21,7 @@ option java_multiple_files = true;
 option java_package = "io.grafeas.v1beta1.vulnerability";
 option objc_class_prefix = "GRA";
 
-// Common Vulnerability Scoring System version 3.
-// For details, see https://www.first.org/cvss/specification-document
+// Deprecated. See CVSS instead
```
aysylu, line 24: could we delete this altogether? v1beta1 implies some backwards-incompatibility is OK and it's better than have confusing similar-looking protos sprinkled through the definitions

proto/v1beta1/vulnerability.proto
```diff
@@ -146,7 +146,11 @@ message Vulnerability {
   // security tracker.
   google.protobuf.Timestamp source_update_time = 6;
 
-  // Next free ID is 7.
+  // The full description of the CVSS.
+  CVSS cvss_v2 = 7;
+  CVSS cvss_v3_0 = 8;
```
aysylu, line 151: let's just update the field type for cvss_v3 instead of creating a new field?

aysylu (changes requested):

proto/v1beta1/vulnerability.proto
nms2140: The new proto lgtm

aysylu (approved):

proto/v1beta1/vulnerability.proto
aysylu: Thank you both!

aysylu:
@hkadakia: this all looks great, merging. Thanks for your contribution, this is a big improvement!

PR #517, opened 2021-11-03
Fixed issue with boolean evaluation on command line value switch truthy

jessesanford:
When trying to run helm install or helm upgrade with the  `--set secret.enabled=true` switch no secrets were being created. I used the same format for evaluation of truthy values as used in the configmap template and it works.


wkozlik (approved):
LGTM. Thank you for this fix.

jessesanford:
> LGTM. Thank you for this fix.

You will need to re-run the scans for this PR as I just signed the CLA

PR #518, opened 2021-11-11
Add CVSS proto to v1 and modify the proto in Occurrences to use this one (this is ok since we have not yet persisted this data and have not released a new version yet).

nms2140:


aysylu (changes requested):
Could you please add to the PR description why it's safe to do this in v1? (because we haven't persisted this data and haven't released the new version, which will be cut in the next couple weeks)

nms2140:
> Could you please add to the PR description why it's safe to do this in v1? (because we haven't persisted this data and haven't released the new version, which will be cut in the next couple weeks)

Done.

aysylu (approved):

PR #521, opened 2021-11-30
Add archive time field to discovery occurrence

titogoog:
Adds an archive time field to the discovery occurrence. The motivation behind this is that occurrence metadata may have an associated retention policy.

e.g. After x days occurrence data related to a discovery is moved into cold storage


proto/v1/discovery.proto
```diff
@@ -80,4 +80,7 @@ message DiscoveryOccurrence {
   
   // The last time this resource was scanned.
   google.protobuf.Timestamp last_scan_time = 5;
+
+  // Output only. The time occurrences related to this discovery occurrence were archived.
```
wkozlik, line 84: Use field behavior OUTPUT_ONLY instead of the comment. https://google.aip.dev/203#output-only

See https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/spdx.proto#L271

proto/v1beta1/discovery.proto
```diff
@@ -84,4 +84,7 @@ message Discovered {
   // details to show to the user. The LocalizedMessage is output only and
   // populated by the API.
   google.rpc.Status analysis_status_error = 4;
+
+  // Output only. The time occurrences related to this discovery occurrence were archived.
```
wkozlik, line 88: Use field behavior OUTPUT_ONLY instead of the comment. https://google.aip.dev/203#output-only

See https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/spdx.proto#L271

wkozlik (changes requested):

proto/v1beta1/discovery.proto
```diff
@@ -84,4 +84,7 @@ message Discovered {
...
+  google.protobuf.Timestamp archive_time = 5;
```
wkozlik, line 89: Let's add this field only in V1 and not in v1beta1.

titogoog: 👍 , done

proto/v1/discovery.proto
titogoog: done

wkozlik (approved):

PR #525, opened 2021-12-06
Update tools/grpc-gateway@v2.7.0

aethanol:
The previous PR to update grpc-gateway missed the go.mod file in tools/.

This PR updates grpc-gateway, changes the import for v1alpha1/grafeas.proto and regenerates (from make test).

Review requested to identify if this breaks anything. The clients will need to be updated with the new swagger.


aysylu (commented):
I'm not entirely sure about the v1alpha1 change, whether that'll break anything. Let's try it and roll this back if we need to. For my own understanding: could you please clarify why this change to v1alpha1 needed to be made?

aethanol:
Ah @aysylu the v1alpha1 import is needed due to requirements of the proto spec. 99% sure that is not an issue since the package gets generated as `grafeas` still.
output from gen:
```
go generate ./proto/v1alpha1
protoc-gen-go: invalid Go import path "grafeas" for "grafeas.proto"

The import path must contain at least one period ('.') or forward slash ('/') character.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.

--go_out: protoc-gen-go: Plugin failed with status code 1.
proto/v1alpha1/generate.go:8: running "../../protoc/bin/protoc": exit status 1
make: *** [generate] Error 1
```

I just updated the go_package to be the equivalent of what is in the v1beta1 import: https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/grafeas.proto#L19 (the package name in the generated files is the same `grafeas`)

aethanol:
I think it's from: https://github.com/protocolbuffers/protobuf-go/releases#v1.26-require-import-path

aysylu:
Thanks @aethanol! I'm guessing the code path with v1alpha1 has not been tested in a long time which is why it's inconsistent with the rest. LGTM, thanks for the fix!

aysylu (approved):

Issue #519, opened 2021-11-23
Add support for CWE field in vulnerability proto message


hkadakia:
Currently we only create a note for a given CVE where the note name is the CVE. There are cases when a particular CVE can have one or more CWEs eg: `CVE-2021-39139`  has CWE `CWE-502` & `CWE-434` and one CWE can be for multiple CVEs.
Ref: [CVE-2021-39139](https://nvd.nist.gov/vuln/detail/CVE-2021-39139)

The latter case can be covered by creating each Note entry for each CVE & having a field within the detail of the note kind which identifies the CWE associated for that CVE.

Since there is a correlation between CVE & CWE creating a different note for each CWE might not be the solution as previously recommended. I would like to propose adding a new field in the vulnerability proto message [here](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/vulnerability.proto#L152) for repeated CWE which can then eventually be trickled down to Note/Occurrence.

Please let me know your thoughts & if any alternate approach than the one suggested above.

Thanks for taking a look.

hkadakia:
@aysylu @nms2140 Any thoughts?

wkozlik:
I think this is a good idea. Would you like to create a PR?

hkadakia:
Thanks. Yes I will raise a PR with the above fix.

aysylu:
@hkadakia sounds good re adding a new field. Should the field be on a Detail or top-level Vulnerability? If different CWEs apply depending on the OS distro, then per-Detail is the best place to do this.

hkadakia:
AFAIK, CWEs are not OS distro dependent and are related to high level CVE/vulnerability. I think adding the field at top-level vulnerability message should be acceptable.

aysylu:
> Afaik, CWEs are not OS distro dependent and are related to high level CVE/vulnerability. I think adding the field at top-level vulnerability message should be acceptable.

@hkadakia Got it. Could we please confirm this with comparison of Debian vs Ubuntu reports of CWEs for a given CVE? I left a comment on the PR, but happy to move the discussion here, up to you.

hkadakia:
Issue resolved as part of #527

PR #527, opened 2021-12-08
add cwe field to vulnerability message in v1beta1

hkadakia:
This PR addresses changes as discussed in https://github.com/grafeas/grafeas/issues/519.


proto/v1beta1/vulnerability.proto
```diff
@@ -149,7 +149,11 @@ message Vulnerability {
   // The full description of the CVSS for version 2.
   CVSS cvss_v2 = 7;
 
-  // Next free ID is 8.
+  // A list of CWE for this vulnerability.
+  // For details, see: https://cwe.mitre.org/index.html
+  repeated string cwe = 8;
```
aysylu, line 154: do we expect the same CWEs to apply on a per-distro basis? or will this vary?

hkadakia: same CWEs should be for all distros.

hkadakia: On checking further, CWEs are just common weakness types, they are general. CVEs are specific vulnerabilities that can affect one or more distros. We could keep CWE as a general field in the vulnerability message as in the PR.

aysylu: Got it, thanks for checking! SGTM. I checked a couple CVEs as well and it appears the CWE association is the same across distros.

aysylu (approved):

aysylu:
Added the co-maintainer, Wiktor, for another pair of eyes. LGTM on my side.

wkozlik (approved):
LGTM

PR #532, opened 2022-01-20
Updates grpc-gateway@v2.7.3

### LINKED ISSUE #379: Swagger spec is broken
## Expected Behavior
Swagger spec should be valid against Swagger spec.

## Actual Behavior
The path parameters are invalid, e.g. https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/swagger/grafeas.swagger.json#L18, leading to errors such as:

    Declared path parameter "name=projects/*/notes/*" needs to be defined as a path parameter at either the path or operation level

The path parameters should simply be enclosed in `{}`.

## Steps to Reproduce the Problem
Validate `grafeas.swagger.json` against swagger spec, e.g. by copying here: https://editor.swagger.io

aethanol:
Fixes #379
grpc-gateway recently updated to replace google.api.http paths with ones that meet the openapi specification.
see: 
https://github.com/grpc-ecosystem/grpc-gateway/pull/2461
https://github.com/grpc-ecosystem/grpc-gateway/pull/2496

One thing to note: grpc-gateway doesn't always (yet?) group operations in the same meaningful way that was intended in the proto file. This impacts the v1alpha1 grafeas.swagger.json in a small way, and was mitigated a bit more by reordering the grafeas.proto file.

### DIFF
--- a/cel/syntax.pb.go
+++ b/cel/syntax.pb.go
@@ -514,7 +514,7 @@ type SourceInfo struct {
 	//
 	// The line number of a given position is the index `i` where for a given
 	// `id` the `line_offsets[i] < id_positions[id] < line_offsets[i+1]`. The
-	// column may be derived from `id_positions[id] - line_offsets[i]`.
+	// column may be derivd from `id_positions[id] - line_offsets[i]`.
 	LineOffsets []int32 `protobuf:"varint,3,rep,packed,name=line_offsets,json=lineOffsets,proto3" json:"line_offsets,omitempty"`
 	// A map from the parse node id (e.g. `Expr.id`) to the character offset
 	// within source.
@@ -587,7 +587,7 @@ type SourcePosition struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// The source location name (e.g. file name).
+	// The soucre location name (e.g. file name).
 	Location string `protobuf:"bytes,1,opt,name=location,proto3" json:"location,omitempty"`
 	// The character offset.
 	Offset int32 `protobuf:"varint,2,opt,name=offset,proto3" json:"offset,omitempty"`
--- a/cel/syntax.pb.go
+++ b/cel/syntax.pb.go
@@ -514,7 +514,7 @@ type SourceInfo struct {
 	//
 	// The line number of a given position is the index `i` where for a given
 	// `id` the `line_offsets[i] < id_positions[id] < line_offsets[i+1]`. The
-	// column may be derivd from `id_positions[id] - line_offsets[i]`.
+	// column may be derived from `id_positions[id] - line_offsets[i]`.
 	LineOffsets []int32 `protobuf:"varint,3,rep,packed,name=line_offsets,json=lineOffsets,proto3" json:"line_offsets,omitempty"`
 	// A map from the parse node id (e.g. `Expr.id`) to the character offset
 	// within source.
@@ -587,7 +587,7 @@ type SourcePosition struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// The soucre location name (e.g. file name).
+	// The source location name (e.g. file name).
 	Location string `protobuf:"bytes,1,opt,name=location,proto3" json:"location,omitempty"`
 	// The character offset.
 	Offset int32 `protobuf:"varint,2,opt,name=offset,proto3" json:"offset,omitempty"`
--- a/proto/v1alpha1/grafeas.proto
+++ b/proto/v1alpha1/grafeas.proto
@@ -1340,26 +1340,6 @@ service Grafeas {
     };
   }
 
-  // Creates a new `Operation`.
-  rpc CreateOperation(CreateOperationRequest)
-  returns (google.longrunning.Operation) {
-    option (google.api.http) = {
-      post: "/v1alpha1/{parent=projects/*}/operations"
-      body: "*"
-    };
-  };
-
-  // Updates an existing operation returns an error if operation
-  //  does not exist. The only valid operations are to update mark the done bit
-  // change the result.
-  rpc UpdateOperation(UpdateOperationRequest)
-  returns (google.longrunning.Operation) {
-    option (google.api.http) = {
-      patch: "/v1alpha1/{name=projects/*/operations/*}"
-      body: "*"
-    };
-  };
-
   // Returns the requested `Note`.
   rpc GetNote(GetNoteRequest) returns (Note) {
     option (google.api.http) = {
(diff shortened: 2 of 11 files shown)


proto/v1alpha1/grafeas.proto
```diff
@@ -1406,6 +1386,26 @@ service Grafeas {
       get: "/v1alpha1/{name=projects/*/notes/*}/occurrences"
     };
   }
+
```
aethanol, line 1389: I reordered the createoperation methods here so it would generate a less weird grafeas.swagger.json file. It's still technically correct, but would generate a somewhat weird client.

I think this tradeoff is worth it since it only impacts v1alpha1.

proto/v1alpha1/swagger/grafeas.swagger.json
```diff
@@ -226,41 +229,97 @@
...
+            "in": "path",
+            "required": true,
+            "type": "string",
+            "pattern": "projects/[^/]+"
+          }
+        ],
+        "tags": [
+          "GrafeasProjects"
+        ]
+      },
```
aethanol, line 274: UpdateOperation is now grouped with the Project methods, this is due to the way grpc-gateway generates swagger "greedily" from google.api.http annotations.

aethanol:
@wkozlik I'm not sure I can request you as a reviewer directly, so mentioning you here. Thanks!

nyc (changes requested):
Thank you for this fix and for contributing the fix to grpc-gateway!

cel/syntax.pb.go
```diff
@@ -514,7 +514,7 @@ type SourceInfo struct {
 	//
 	// The line number of a given position is the index `i` where for a given
 	// `id` the `line_offsets[i] < id_positions[id] < line_offsets[i+1]`. The
-	// column may be derived from `id_positions[id] - line_offsets[i]`.
+	// column may be derivd from `id_positions[id] - line_offsets[i]`.
```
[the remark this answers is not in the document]
aethanol: yeah can do that: the typo is from https://github.com/googleapis/googleapis/blob/master/google/api/expr/v1alpha1/syntax.proto#L295

aethanol:
Follow up question @nyc do you know why were are generating a specific version of syntax in cel? https://github.com/grafeas/grafeas/blob/master/cel/generate.go#L3 

Seems like the only meaningful difference is adding macro_calls?
`git diff e266a1ca806aac193ee3ed5d6d2031bea351546e HEAD google/api/expr/v1alpha1/syntax.proto`
```
-// Copyright 2018 Google LLC.
+// Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -11,7 +11,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//

 syntax = "proto3";

@@ -45,24 +44,21 @@ message ParsedExpr {
 // operators with the exception of the '.' operator are modelled as function
 // calls. This makes it easy to represent new operators into the existing AST.
 //
-// All references within expressions must resolve to a
-// [Decl][google.api.expr.v1alpha1.Decl] provided at type-check for an
-// expression to be valid. A reference may either be a bare identifier `name` or
-// a qualified identifier `google.api.name`. References may either refer to a
-// value or a function declaration.
+// All references within expressions must resolve to a [Decl][google.api.expr.v1alpha1.Decl] provided at
+// type-check for an expression to be valid. A reference may either be a bare
+// identifier `name` or a qualified identifier `google.api.name`. References
+// may either refer to a value or a function declaration.
 //
 // For example, the expression `google.api.name.startsWith('expr')` references
-// the declaration `google.api.name` within a
-// [Expr.Select][google.api.expr.v1alpha1.Expr.Select] expression, and the
-// function declaration `startsWith`.
+// the declaration `google.api.name` within a [Expr.Select][google.api.expr.v1alpha1.Expr.Select] expression, and
+// the function declaration `startsWith`.
 message Expr {
   // An identifier expression. e.g. `request`.
   message Ident {
     // Required. Holds a single, unqualified identifier, possibly preceded by a
     // '.'.
     //
-    // Qualified names are represented by the
-    // [Expr.Select][google.api.expr.v1alpha1.Expr.Select] expression.
+    // Qualified names are represented by the [Expr.Select][google.api.expr.v1alpha1.Expr.Select] expression.
     string name = 1;
   }

@@ -103,7 +99,7 @@ message Expr {

   // A list creation expression.
   //
-  // Lists may either be homogenous, e.g. `[1, 2, 3]`, or heterogenous, e.g.
+  // Lists may either be homogenous, e.g. `[1, 2, 3]`, or heterogeneous, e.g.
   // `dyn([1, 'hello', 2.0])`
   message CreateList {
     // The elements part of the list.
@@ -239,8 +235,7 @@ message Expr {
 // primitives.
 //
 // Lists and structs are not included as constants as these aggregate types may
-// contain [Expr][google.api.expr.v1alpha1.Expr] elements which require
-// evaluation and are thus not constant.
+// contain [Expr][google.api.expr.v1alpha1.Expr] elements which require evaluation and are thus not constant.
 //
 // Examples of literals include: `"hello"`, `b'bytes'`, `1u`, `4.2`, `-2`,
 // `true`, `null`.
@@ -292,16 +287,27 @@ message SourceInfo {
   // `acme/app/AnvilPolicy.cel`.
   string location = 2;

-  // Monotonically increasing list of character offsets where newlines appear.
+  // Monotonically increasing list of code point offsets where newlines
+  // `\n` appear.
   //
   // The line number of a given position is the index `i` where for a given
   // `id` the `line_offsets[i] < id_positions[id] < line_offsets[i+1]`. The
   // column may be derivd from `id_positions[id] - line_offsets[i]`.
   repeated int32 line_offsets = 3;

-  // A map from the parse node id (e.g. `Expr.id`) to the character offset
-  // within source.
+  // A map from the parse node id (e.g. `Expr.id`) to the code point offset
+  // within the source.
   map<int64, int32> positions = 4;
+
+  // A map from the parse node id where a macro replacement was made to the
+  // call `Expr` that resulted in a macro expansion.
+  //
+  // For example, `has(value.field)` is a function call that is replaced by a
+  // `test_only` field selection in the AST. Likewise, the call
+  // `list.exists(e, e > 10)` translates to a comprehension expression. The key
+  // in the map corresponds to the expression id of the expanded macro, and the
+  // value is the call `Expr` that was replaced.
+  map<int64, Expr> macro_calls = 5;
 }

 // A specific position in source.
@@ -309,7 +315,7 @@ message SourcePosition {
   // The soucre location name (e.g. file name).
   string location = 1;

-  // The character offset.
+  // The UTF-8 code unit offset.
   int32 offset = 2;

   // The 1-based index of the starting line in the source text
   ```

nyc:
@aethanol I'm not quite sure as I don't have any context about that directory, so I would have to take a deeper look. Do you use the generated file in that directory at all?

nyc (approved):

nyc:
@wkozlik LGTM to merge.

wkozlik:
Thank you for this PR!

PR #539, opened 2022-03-16
Add a "benchmark_document" field to the compliance proto.

erikvarga:
This field will be used to store a reference to the document that defines the given benchmark, making it easier for product teams to look up information about compliance findings. For most OSes, each CPE URI will have a corresponding benchmark document.


wkozlik (approved):
LGTM

PR #541, opened 2022-04-06
Adding a way to indicate the location at which a package was found on PackageIssue

nms2140:


proto/v1/vulnerability.proto
```diff
@@ -207,6 +207,16 @@ message VulnerabilityOccurrence {
     // when that is available and note provider assigned severity when it is not
     // available.
     grafeas.v1.Severity effective_severity = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+    // Indicates the lcation at which a package was found.
+    message FileLocation {
```
wkozlik, line 212: Would it make sense to define the message FileLocation outside of VulnerabilityOccurrence so it could be used in other protos, for example SPDX PackageInfoOccurrence?

proto/v1/vulnerability.proto
```diff
@@ -207,6 +207,16 @@ message VulnerabilityOccurrence {
     // when that is available and note provider assigned severity when it is not
     // available.
     grafeas.v1.Severity effective_severity = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
+
+    // Indicates the lcation at which a package was found.
```
wkozlik, line 211: nit: typo in "location"

proto/v1/vulnerability.proto
nms2140: I have moved it to common.proto, so people will not need to include vulnerability.proto if they want to use FileLocation.

proto/v1/vulnerability.proto
nms2140: done.

wkozlik (approved):
LGTM

PR #542, opened 2022-04-07
Regenerates proto files and cleans up Swagger for `v1beta1`

nyc:
The changes to `cel` and `v1` come from re-running `make generate` on
the current protos.

The generator for `v1beta1` is updated to do a few things:
* Generate a merged Swagger file that includes both the `Grafeas` and
  `Projects` services.
* Stops generating Swagger files for protos that do not define any
  services. This is pointless since required definitions are already
  recursively put into the Swagger file for the service where it is
  needed.

Running this generator removes all of the unnecessary Swagger files from
`v1beta1` and creates the merged Swagger file for `v1beta1`.


cel/syntax.pb.go
```diff
@@ -514,7 +514,7 @@ type SourceInfo struct {
 	//
 	// The line number of a given position is the index `i` where for a given
 	// `id` the `line_offsets[i] < id_positions[id] < line_offsets[i+1]`. The
-	// column may be derived from `id_positions[id] - line_offsets[i]`.
+	// column may be derivd from `id_positions[id] - line_offsets[i]`.
```
nyc, line 517: Unfortunate typos, but we should probably just let this through since anyone who tries to generate in the future will keep regenerating this change.

wkozlik (approved):
LGTM

PR #544, opened 2022-04-08
Regenerate for `v1` API and add Swagger for `v1`

nyc:


nyc:
Changes in the `pb.go` files are from @nms2140 changes in #541

nms2140 (approved):

PR #548, opened 2022-05-05
Make cvss proto message compatible with both v3 and v2.

zhaoyonghe:
The current v1beta1 CVSS message only supports v3. We cannot correctly fill in [the cvss_v2 field in Vulnerability](https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/vulnerability.proto#L150). It cannot represent a CVSS v2 message that, for example, includes `IMPACT_PARTIAL`.


wkozlik (approved):
LGTM

wkozlik:
Thank you for this fix.

PR #549, opened 2022-05-06
Fix go1.18 vet error: (*testing.common).Error call has possible formatting directive.

zhaoyonghe:
`go vet` (go1.18) fails currently.
```
> go version
go version go1.18 darwin/amd64
> go vet ./...
# github.com/grafeas/grafeas/go/filtering/common
go/filtering/common/source_test.go:128:3: (*testing.common).Error call has possible formatting directive %d
go/filtering/common/source_test.go:130:3: (*testing.common).Error call has possible formatting directive %s
```
This pr fixes it.


nyc (approved):

nyc:
Good catch and thank you!

PR #550, opened 2022-05-10
Add support for SLSA v0.2 to the intoto statement.

sherzberg-1:
The main changes are listed in https://slsa.dev/provenance/v0.2 in the Change History section. We also changed the Any's to Struct's because they're defined as JSON objects. We prefer Struct over any, per https://slsa.dev/provenance/v0.2. To convert from an arbitrary proto to a proto Struct, on can use https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewValue to get a Value, and then https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#Value.GetStructValue to get the Struct from there.

### DIFF
--- a/proto/v1/slsa_provenance_0_2.proto
+++ b/proto/v1/slsa_provenance_0_2.proto
@@ -1,0 +1,82 @@
+// Copyright 2021 The Grafeas Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


proto/v1/slsa_provenance_0_2.proto
```diff
@@ -0,0 +1,82 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package grafeas.v1;
+
+import "google/protobuf/struct.proto";
```
wkozlik, line 19: Move imports below options.

wkozlik: Could you include re-generated Go code and swagger?

PR #552, opened 2022-05-23
Change the type of digest_value to bytes to store raw digests instead of encoded values

vincentyl:
Change the type of digest_value to bytes to store raw digests instead of encoded values


proto/v1/common.proto
```diff
@@ -144,9 +144,8 @@ message Digest {
   // `SHA1`, `SHA512` etc.
   string algo = 1;
 
-  // Value of the digest encoded.
-  // For example: SHA512 - base64 encoding, SHA1 - hex encoding.
-  string digest_value = 2;
+  // Value of the digest.
+  bytes digest_value = 2;
```
nyc, line 148: While `string` and `bytes` are technically compatible in protos, my concern is that any generated clients will change from `string -> []byte`, which is potentially a breaking change.

If you need to store un-encoded bytes, I would suggest adding a new field `digest_bytes` with a new tag value. If you need to guarantee exclusivity of the field, you can add it as part of a `oneof` field.

proto/v1beta1/common.proto
```diff
@@ -141,7 +141,6 @@ message Digest {
   // `SHA1`, `SHA512` etc.
   string algo = 1;
 
-  // Value of the digest encoded.
-  // For example: SHA512 - base64 encoding, SHA1 - hex encoding.
-  string digest_value = 2;
+  // Value of the digest.
+  bytes digest_value = 2;
```
nyc, line 145: See other comment in `v1`.

nyc (changes requested):

proto/v1/common.proto
nyc: Thoughts @wkozlik?

wkozlik: This field has been added very recently, and hasn't been released yet. For that reason, this change could be treated as a correction, and doesn't need to follow the "non-breaking" change routine.

vincentyl: This is intended to be a  breaking change since it's a semantic fix that's not supposed to be backward compatible.

Made the changes correspondingly. PTAL.

nyc (approved):

PR #554, opened 2022-06-14
Add SLSA 0.2 converter

chuangw6:
`InTotoStatement_SlsaProvenanceZeroTwo` predicate in grafeas
is wrapped as `oneof` predicates. That means a JSONified intoto statement
with slsa 0.2 predicate cannot be directly unmarshalled to grafeas
intoto statement.

In this commit, we add a conversion layer that converts a JSONified intoto
statement with slsa 0.2 predicate to the intoto statement compliant to
grafeas.


chuangw6:
cc @wlynch @sherzberg-1 @wkozlik @r2wend2

chuangw6:
cc @wlynch @sherzberg-1 @wkozlik @r2wend2

wlynch (changes requested):

utils/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,124 @@
+package utils
```
wkozlik, line 1: Could you move the `utils` folder under `/go` ?

wkozlik (changes requested):

utils/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,124 @@
...
+
+	pb "github.com/grafeas/grafeas/proto/v1/grafeas_go_proto"
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// ConvertSLSAZeroTwo function converts the raw intoto statement with SLSA 0.2 predicate to the
+// grafeas proto InTotoStatement where slsa 0.2 predicate is oneof the predicate types
+func ConvertSLSAZeroTwo(rawPayload []byte) (*pb.InTotoStatement, error) {
```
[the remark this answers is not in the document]
chuangw6: Do you mean convert from slsa0.2 proto to slsa0.2 intoto statement only? What are the use cases for it?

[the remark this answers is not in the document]
chuangw6: good point. sgtm! The step from `[]byte` to `*intoto.ProvenanceStatement` can be done in chains!

utils/slsa_zero_two_converter.go
chuangw6: sure! sgtm!

utils/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func convertMetaData(input intoto.ProvenanceStatement) *pb.SlsaProvenanceZeroTwo_SlsaMetadata {
+	if input.Predicate.Metadata == nil {
+		return nil
+	}
+
+	return &pb.SlsaProvenanceZeroTwo_SlsaMetadata{
+		BuildInvocationId: input.Predicate.Metadata.BuildInvocationID,
+		BuildStartedOn:    timestamppb.New(*input.Predicate.Metadata.BuildStartedOn),
```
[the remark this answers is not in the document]
chuangw6: Good catch!!

utils/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+
+	environment, err := convertEmptyInterfaceToProtoStruct(input.Predicate.Invocation.Environment)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.InTotoStatement_SlsaProvenanceZeroTwo{
+		SlsaProvenanceZeroTwo: &pb.SlsaProvenanceZeroTwo{
```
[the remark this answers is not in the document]
chuangw6: Yes there are more differences so we need to manually convert them.
Specifically, [some json field names of predicate in intoto-golang](https://github.com/in-toto/in-toto-golang/blob/07976bf395d5fbc5030b4a576579ee70cd64cf88/in_toto/slsa_provenance/v0.2/provenance.go#L11-L18) are camelCase, whereas [the corresponding fields in grafeas proto](https://github.com/grafeas/grafeas/blob/bc2e0363cdefcb36f4f5fa20bf45e7ff9f588030/proto/v1/grafeas_go_proto/slsa_provenance_zero_two.pb.go#L39-L50) are underscore_separated_names since grafeas proto is following [proto style guide](https://developers.google.com/protocol-buffers/docs/style#message_and_field_names) to use unsersore_separated_names.

Other than that, some fields in grafeas are of type `*_struct.Struct`, and the correspondings are of type `interface{}` i.e. `BuildConfig` in predicate.

go/utils/intoto/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return results
+}
+
+func convertEmptyInterfaceToProtoStruct(input interface{}) (*structpb.Struct, error) {
+	if input == nil {
+		return nil, nil
+	}
+
+	data, err := structpb.NewStruct(input.(map[string]interface{}))
```
wlynch, line 88: I think this panics if the incoming type is not `map[string]interface{}` - you may need to serialize the input -> map[string]interface{} first to make sure this works for all types.

go/utils/intoto/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,131 @@
...
+
+	pb "github.com/grafeas/grafeas/proto/v1/grafeas_go_proto"
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// SLSAZeroTwoToProto function converts the intoto statement with SLSA 0.2 predicate to the
+// grafeas proto InTotoStatement where slsa 0.2 predicate is oneof the predicate types
+func SLSAZeroTwoToProto(in *intoto.ProvenanceStatement) (*pb.InTotoStatement, error) {
```
wlynch, line 15: bikeshed: 
```suggestion
func ToSLSAZeroTwoProto(in *intoto.ProvenanceStatement) (*pb.InTotoStatement, error) {
```

`SLSAZeroTwoToProto` can read as `SLSAZeroTwo` -> `Proto`, which isn't what this is doing.

go/utils/intoto/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, subject := range input {
+		results = append(results, &pb.Subject{
+			Name:   subject.Name,
+			Digest: subject.Digest,
+		})
+	}
+	return results
+}
+
+func convertEmptyInterfaceToProtoStruct(input interface{}) (*structpb.Struct, error) {
```
wlynch, line 83: bikeshed: 
```suggestion
func toProtoStruct(input interface{}) (*structpb.Struct, error) {
```

go/utils/intoto/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,131 @@
...
+
+	data, err := structpb.NewStruct(input.(map[string]interface{}))
+	if err != nil {
+		return nil, fmt.Errorf("Unable to convert the interface{} to *structpb.Struct: %v", err)
+	}
+
+	return data, nil
+}
+
+func convertMetaData(in *intoto.ProvenanceStatement) *pb.SlsaProvenanceZeroTwo_SlsaMetadata {
```
wlynch, line 96: nit:
```suggestion
func convertMetadata(in *intoto.ProvenanceMetadata) *pb.SlsaProvenanceZeroTwo_SlsaMetadata {
```

go/utils/intoto/slsa_zero_two_converter.go
```diff
@@ -0,0 +1,131 @@
...
+		Completeness: &pb.SlsaProvenanceZeroTwo_SlsaCompleteness{
+			Parameters:  in.Predicate.Metadata.Completeness.Parameters,
+			Environment: in.Predicate.Metadata.Completeness.Environment,
+			Materials:   in.Predicate.Metadata.Completeness.Materials,
+		},
+		Reproducible: in.Predicate.Metadata.Reproducible,
+	}
+}
+
+func convertMaterials(in *intoto.ProvenanceStatement) []*pb.SlsaProvenanceZeroTwo_SlsaMaterial {
```
wlynch, line 114: nit: 
```suggestion
func convertMaterials(in []*intoto.ProvenanceMaterials) []*pb.SlsaProvenanceZeroTwo_SlsaMaterial {
```

go/utils/intoto/slsa_zero_two_converter.go
wlynch: (we should also add a test for this)

utils/slsa_zero_two_converter.go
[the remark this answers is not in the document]
wlynch: I suspect that you'll want to be able to convert both ways - i.e. intoto->grafeas and grafeas->intoto

If you want to hold off on the latter that's fine though.

utils/slsa_zero_two_converter.go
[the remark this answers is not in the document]
wlynch: IIRC, proto3 will accept both camel and underscore style names during unmarshalling, right? so these are only issues if we're going from grafeas -> intoto?

go/utils/intoto/slsa_zero_two_converter.go
chuangw6: aha. Let me just use `ToProto` then.

utils/slsa_zero_two_converter.go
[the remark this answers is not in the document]
chuangw6: Just tried to do direct unmarshal on the predicate part. It skips all the underscore fields i.e.  `build_config`, `build_type`, and some metadata fields named with underscore style in proto.

go/utils/intoto/slsa_zero_two_converter.go
chuangw6: Since [structpb.struct](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#Struct) only expects a map rather than any other type i.e string, slice, the actual type of the incoming interface{} has to be a map right?

Just changed `toProtoStruct` function in which we do marshal -> unmarshal to structpb.struct so that error will be returned instead of panic when the incoming type is not map[string]interface{}. Let me know what you think.

chuangw6: btw, I am wondering if there is a proto type that can be autotranslated to go's `interface{}` type?

cc @sherzberg-1 since [build_config](https://github.com/grafeas/grafeas/blob/bc2e0363cdefcb36f4f5fa20bf45e7ff9f588030/proto/v1/slsa_provenance_zero_two.proto#L79), [parameters and environment](https://github.com/grafeas/grafeas/blob/bc2e0363cdefcb36f4f5fa20bf45e7ff9f588030/proto/v1/slsa_provenance_zero_two.proto#L47-L48) fields are designed to be structpb.Struct type.

utils/slsa_zero_two_converter.go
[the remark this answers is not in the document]
chuangw6: sg, let's add that later.

wlynch (approved):

PR #559, opened 2022-09-15
Added cvss_version field to vulnerability Note and Occurrence

wkozlik:
This is to provide information which of CVSS v3 or CVSS v2 were used to populate fields: cvss_score and severity.


proto/v1/cvss.proto
```diff
@@ -157,3 +157,10 @@ message CVSS {
     IMPACT_NONE = 3;
   }
 }
+
+// CVSS Version.
+enum CVSSVersion {
+  CVSS_VERSION_UNSPECIFIED = 0;
+  CVSS_VERSION_2 = 2;
```
sherzberg-1, line 164: Is there a reason you're starting from 2? This is just a proto field, so is there any place this value would get exposed as 2?

proto/v1/vulnerability.proto
```diff
@@ -147,6 +147,11 @@ message VulnerabilityNote {
   // upstream timestamp from the underlying information source - e.g. Ubuntu
   // security tracker.
   google.protobuf.Timestamp source_update_time = 6;
+
+  // Output only. CVSS version used to populate cvss_score and severity.
```
sherzberg-1, line 151: Why output only? On the note we store/populate it.

proto/v1/cvss.proto
wkozlik: Thanks for catching this. Corrected.

proto/v1/vulnerability.proto
wkozlik: Done.

sherzberg-1 (approved):

PR #565, opened 2023-01-02
Use ANTLR v4

zregvart:
Using `github.com/antlr/antlr4` is problematic when combined with
dependencies that use the newer
`github.com/antlr/antlr4/runtime/Go/antlr/v4`. Following the ANTLR v4
golang guide[1] this upgrades to the latest v4 release.

[1] https://github.com/antlr/antlr4/blob/master/doc/go-target.md#upgrading-to-v4-from-the-default-path


wkozlik (approved):
Thank you for your contribution, and I apologize for the delay.

PR #575, opened 2023-03-07
Remove remediation_uri field, change publisher.context to publisher_namespace.

yeshwanth1993:
Remove remediation_uri field, change publisher.context to publisher_namespace.


wkozlik (approved):
LGTM

PR #576, opened 2023-03-07
Move vex to v1 api from v1beta1

yeshwanth1993:


wkozlik (changes requested):
We need to add the vulnerability_assessment to V1 grafeas.proto, similarly to https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/grafeas.proto#L351, and VULNERABILITY_ASSESSMENT NoteKind like in https://github.com/grafeas/grafeas/blob/master/proto/v1beta1/common.proto#L53.

yeshwanth1993:
PR is updated. I have added it to the note kind and to the grafeas.proto/Note

wkozlik (approved):
LGTM

PR #578, opened 2023-03-17
Add SBOM Reference to v1b1

sherzberg-1:


proto/v1beta1/common.proto
```diff
@@ -146,3 +146,14 @@ message Digest {
   // Value of the digest.
   bytes digest_bytes = 2;
 }
+
+message IntotoStatement {
```
wkozlik, line 150: Is this the same type as defined in V1? https://github.com/grafeas/grafeas/blob/master/proto/v1/intoto_statement.proto#L32

How are we going to reconcile them when we promote SBOM to V1?

wkozlik (changes requested):

proto/v1beta1/common.proto
sherzberg-1: Unfortunately that kind is not going to work well here. Specifically, the oneof, when jsonified, gets the field name of the inner field and not the name of the oneof field itself (predicate), meaning its not actually compliant with the intoto format. So we won't be using that proto at all for SBOM.

When we move this to v1, it will be the same thing as I'm defining here, in v1b1. I've manually confirmed (and will add in unit tests to confirm) that jsonifying this proto returns a body that's in spec to intoto.

wkozlik (approved):
LGTM

PR #582, opened 2023-03-29
add enum for grafeas v1 cvss to support both cvss v2 and v3

hanqiuzh:
The Grafeas v1 CVSS class was designed for both CVSS v2 and v3. 
See the [comment](https://github.com/grafeas/grafeas/blob/master/proto/v1/cvss.proto#L90-L91) above the CVSS definition:
> // This is a message we will try to use for storing various versions of CVSS
> // rather than making a separate proto for storing a specific version.

However, Currently it is missing some enum values of CVSS v2:
- ATTACK_COMPLEXITY_MEDIUM
- IMPACT_PARTIAL
- IMPACT_COMPLETE

Also ran `make test` to generate the protos.


wkozlik (approved):
LGTM

PR #588, opened 2023-06-09
SBOM

sherzberg-1:
1. Update notekind proto values.
2. Add InTotoSlsaProvenanceV1 to v1beta1.

We decided to copy the provenance over to Grafeas, instead of having it autoimport, so that we can control any changed that may happen, and import them if/when we want.


sherzberg-1:
Note that we will not be moving this to v1 until https://github.com/in-toto/attestation/issues/242  is resolved, as we do not want to make incompatible changes outside of v1beta1.

wkozlik (approved):
LGTM

PR #591, opened 2023-06-16
Change Struct to Value in ResourceDescriptor.

sherzberg-1:
See https://github.com/in-toto/attestation/issues/242 and https://github.com/slsa-framework/slsa/issues/875 for the conversation on this topic.


hmeng22 (approved):
LGTM

PR #594, opened 2023-06-28
Don't use doubly nested wrapper for intoto slsa statement

sherzberg-1:


hmeng22 (approved):
LGTM, those should be top-level fields in the InTotoSlsaProvenanceV1

PR #606, opened 2023-10-06
Add an optional impact field to ComplianceNote

dguerri:
The impact field describes the potential impact of the proposed remediation, or more generally, the potential downsides and caveats caused by fulfilling a particular compliance requirement. This field is present in CIS benchmarks spec, especially for Level 2 benchmarks.

The added field is optional and doesn't add any additional requirement for Grafeas itself, being for informational purposes only.


wkozlik (approved):
LGTM

PR #610, opened 2024-02-21
Add locations to Read methods

hmeng22:
Add additional bindings in Read methods, to add an optional `locations` parameter


wkozlik (approved):
LGTM

PR #613, opened 2024-04-17
Add the compliance version field to the compliance occurrences too.

erikvarga:
This will allow scanners like Localtoast to mark what CPE / benchmark document it was using to perform a given scan.


wkozlik (approved):
LGTM

PR #617, opened 2024-12-07
Remove references to the deprecated Grafeas reference server image.

rhdesmond:
The Grafeas reference server images aren't being used and are not worth the maintenance and security compliance efforts.  Remove references in the docs to the public, published Grafeas server reference image.  The next step toward deprecation is to delete the public images.


wkozlik (approved):
LGTM

PR #621, opened 2025-02-06
[Proto Update] Add the LayerDetails proto to the PackageIssue message

mleyvajr100:
Added the `layer_details` field to the `PackageIssue` message in the `v1` and `v1beta1` APIs. The new `LayerDetails` proto will be used to augment vulnerability occurrences with layer metadata. This should help users find which layer in a container image was responsible for introducing a vulnerability.


proto/v1/vulnerability.proto
```diff
@@ -219,6 +219,10 @@ message VulnerabilityOccurrence {
 
     // The location at which this package was found.
     repeated FileLocation file_location = 10;
+
+    // Each package should have its own layer metadata (that is, information
+    // from the origin layer of the package).
+    LayerDetails layer_details = 11;
```
wkozlik, line 225: Wouldn't make sense to include the layer_details inside file_location?  It is because a single package can be present in multiple file locations, possibly in different layers.

mleyvajr100: That was the original plan, but it's a bit weird to have multiple `layer_details` for a package issue. We want to be able to tell users which layer they should focus on remediating a certain vulnerability. If the package comes from multiple files (potentially across multiple layers), then we should tell the user that the vulnerability came from the most recent layer. Below is an example of why'd we want to associate a package issue with a single layer:

Layer 1: Base image has package X version Y (vulnerable to CVE).
Layer 2: Platform team of some company patches a vulnerability by overwriting the file with package X version Z (not vulnerable).
Layer 3: Application team accidentally overwrites package X with version Y (vulnerable) again.

In this case, we'd want to say that the vulnerability originated in just Layer 3 (not Layer 1 or Layer 1 and Layer 3).

wkozlik: This is true when the single package file is overwritten in the next layer. But the `file_location` is not concerned about which layer the package is found in, but rather distinct locations in the file system.

For example, if package A is found in two different locations: `/dir_a/a.jar` in layer 1, and in `/dir_b/a.jar` in layer 2, which of these two files the `layer_details` should correspond to?

mleyvajr100:
Discussed offline. Decided to place the `LayerDetails` proto in the `FileLocation` message.

PR #622, opened 2025-02-11
[Proto Update] Add LayerDetails proto to the FileLocation message

mleyvajr100:
This is a follow-up PR that instead of storing the `LayerDetails` proto directly in the `PackageIssue` message, we will instead store it in the `FileLocation` message since packages from different files should have their own layer details.


mleyvajr100:
Reference PR: https://github.com/grafeas/grafeas/pull/621

wkozlik (approved):
LGTM. Thank you!

PR #623, opened 2025-02-12
[Fix] Use int32 for num_layers field for the BaseImage proto

mleyvajr100:
Change the `num_layers` field in the `BaseImage` proto from type string to type int32.


wkozlik (approved):
LGTM

PR #624, opened 2025-02-12
[Proto Update] Rename "num_layers" field in the "BaseImage" proto to "layer_count" 

mleyvajr100:
The Google API style guide suggests using "{entity}_count" rather than "num_{entity}".


wkozlik (approved):
LGTM

PR #628, opened 2025-04-29
Add Secret Kind to Grafeas v1.

hanqiuzh:


proto/v1/common.proto
```diff
@@ -49,6 +49,8 @@ enum NoteKind {
   VULNERABILITY_ASSESSMENT = 11;
   // This represents an SBOM Reference.
   SBOM_REFERENCE = 12;
+  // This represents a secret.
+  SECRET = 18;
```
wkozlik, line 53: Why not the next available value, i.e 13?

wkozlik (changes requested):

proto/v1/common.proto
hanqiuzh: good catch. will have another pr to fix

hanqiuzh: Fixed in https://github.com/grafeas/grafeas/pull/629

PR #640, opened 2025-08-28
add secret in v1beta1

hanqiuzh:


hmeng22 (approved):
LGTM

hanqiuzh:
Good question. I don't plan adding the data field in v1beta1. 
Not sure about if we will back-port layerDetails in v1beta1, but I guess that can be in a future CL if we want to add it.

PR #657, opened 2026-06-15
Add exploit maturity to CVSS proto.

hanqiuzh:
Mainly defined values to support threat metrics in CVSS v4.
Temporal metrics in v3 will not be back-ported (although it's possible we can reuse exploit_maturity for `Exploit Code Maturity (E)` in v3 if we decide to back-port v3 temporal metrics).


wkozlik (approved):
LGTM

